PR #1, opened 2019-06-07
Taskcluster CI/CD

La0:
Tasks to implement:

- [x] Check Python code
- [x] Build docker image
- [x] Build web app
- [x] Deploy docker image on dockerhub (on branches: `testing, staging, production`)
- [x] Update taskcluster hook with new docker image (on branches: `testing, staging, production`)
- [x] Deploy web app as S3 files (on branches: `testing, staging, production`)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -83,3 +83,36 @@ tasks:
         description: Build web single page application
         owner: bastien@mozilla.com
         source: https://github.com/mozilla/code-review
+
+    - taskId: {$eval: as_slugid("bot_build")}
+      created: {$fromNow: ''}
+      deadline: {$fromNow: '1 hour'}
+      provisionerId: aws-provisioner-v1
+      workerType: releng-svc
+      dependencies:
+        - {$eval: as_slugid("bot_checks")}
+      payload:
+        capabilities:
+          privileged: true
+        maxRunTime: 3600
+        image: mozilla/taskboot:0.1.6
+        env:
+          GIT_REPOSITORY: ${repository}
+          GIT_REVISION: ${head_rev}
+        command:
+          - taskboot
+          - build
+          - --write
+          - /images/code-review-bot.tar
+        artifacts:
+          public/bugbug:
+            expires: {$fromNow: '2 weeks'}
+            path: /images
+            type: directory
+      scopes:
+        - docker-worker:capability:privileged
+      metadata:
+        name: Code Review Bot docker build
+        description: Build docker image of code review bot
+        owner: bastien@mozilla.com
+        source: https://github.com/mozilla/code-review
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -32,7 +32,7 @@ tasks:
         else: 'dev'
       else: 'dev'
 
-    taskboot_image: "mozilla/taskboot@sha256:34cf551e94e53f8695187dcb2a0c602e1c605750bda433c1968c89089ed408ad"
+    taskboot_image: "mozilla/taskboot@sha256:0.1.7"
   in:
     - taskId: {$eval: as_slugid("bot_check_lint")}
       provisionerId: aws-provisioner-v1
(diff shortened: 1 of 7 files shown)


marco-c:
r+ but it'd be nice to trigger a Taskcluster build before merging to ensure it works!

La0:
Oh it's far from finished, it was to test the Taskcluster trigger... I've updated the description to list all the needed tasks.

marco-c:
> Oh it's far from finished, it was to test the Taskcluster trigger...

Yeah, but we could also merge one thing at a time and add more tasks in separate PRs.

La0:
Images available on https://hub.docker.com/r/mozilla/code-review/tags

La0:
[First green task](https://tools.taskcluster.net/groups/VtSvj6wySYCjsODKOFy2ww/tasks/VtSvj6wySYCjsODKOFy2ww/runs/0/logs/public%2Flogs%2Flive.log) on testing !

.taskcluster.yml
```diff
@@ -0,0 +1,227 @@
...
+      workerType: github-worker
+      created: {$fromNow: ''}
+      deadline: {$fromNow: '1 hour'}
+      extra:
+        github:
+          events:
+            - pull_request.opened
+            - pull_request.reopened
+            - pull_request.synchronize
+            - push
```
marco-c: I think this extra part is no longer needed with version 1 of .taskcluster.yml

.taskcluster.yml
```diff
@@ -0,0 +1,227 @@
...
+      payload:
+        maxRunTime: 3600
+        image: python:3-alpine
+        command:
+          - sh
+          - -lxce
+          - "apk add git --quiet &&
+            git clone --quiet ${repository} /src && cd /src/bot && git checkout ${head_rev} -b checks &&
```
marco-c: Nit: you could remove the detached head advice which is just noise: https://github.com/mozilla/bugbug/blob/37f93de33b0bb6523c51b6a65b770998921fef9d/.taskcluster.yml#L44

.taskcluster.yml
```diff
@@ -0,0 +1,227 @@
...
+            pip install . && pip install -r requirements-dev.txt &&
+            flake8 && pytest -v"
```
marco-c: Nit: I'd split linting and testing in two separate tasks

.taskcluster.yml
```diff
@@ -0,0 +1,227 @@
...
+          - build
+          - --image
+          - mozilla/code-review
+          - --tag
+          - "${channel}"
+          - --tag
+          - "${head_rev}"
+          - --write
+          - /bot.tar
+          - bot/Dockerfile
```
marco-c: Is this going to write an image with two tags?

.taskcluster.yml
```diff
@@ -0,0 +1,227 @@
...
+      then: ${event.pull_request.head.repo.html_url}
+      else: ${event.repository.html_url}
+
+    channel:
+      $if: 'tasks_for == "github-push"'
+      then:
+        $if: 'event.ref in ["refs/heads/testing", "refs/heads/staging", "refs/heads/production"]'
+        then: ${event.ref[11:]}
+        else: 'dev'
+      else: 'dev'
```
marco-c: I wonder if we couldn't just live with two deployment channels now.
In release-services, we created 'testing' simply because there were too many projects and there was always a queue for pushing things to staging.

La0: Ok for testing + production

.taskcluster.yml
La0: Yeap

La0: There may be a bug on taskboot, as i only see the hashes on the repo, not the `testing` tag

marco-c: Maybe we need something like https://github.com/mozilla/task-boot/pull/38 for the build command too?

La0: I think it should be in the `push-artifact` command to iterate over tags and push all of them https://github.com/mozilla/task-boot/blob/master/taskboot/push.py#L52 instead of just one

PR #36, opened 2019-06-13
bot: Support NSS tasks.

La0:
Needed to support https://github.com/mozilla/release-services/pull/2180 and [Bug 1559095](https://bugzilla.mozilla.org/show_bug.cgi?id=1559095)

Could also be used to update the variable name on MC

:warning: The `reliability` key in Coverity payload is now optional.

### DIFF
--- a/bot/code_review_bot/config.py
+++ b/bot/code_review_bot/config.py
@@ -16,7 +16,8 @@ import structlog
 import yaml
 
 PROJECT_NAME = 'code-review-bot'
-REPO_TRY = b'https://hg.mozilla.org/try'
+REPO_GECKO_TRY = 'https://hg.mozilla.org/try'
+REPO_NSS_TRY = 'https://hg.mozilla.org/projects/nss-try'
 RAW_FILE_URL = 'https://hg.mozilla.org/mozilla-central/raw-file/tip/{}'
 
 logger = structlog.get_logger(__name__)
--- a/bot/code_review_bot/revisions.py
+++ b/bot/code_review_bot/revisions.py
@@ -13,7 +13,8 @@ from parsepatch.patch import Patch
 
 from code_review_bot import Issue
 from code_review_bot import stats
-from code_review_bot.config import REPO_TRY
+from code_review_bot.config import REPO_GECKO_TRY
+from code_review_bot.config import REPO_NSS_TRY
 from code_review_bot.config import settings
 from code_review_bot.tools.taskcluster import create_blob_artifact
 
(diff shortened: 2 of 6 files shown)


bot/code_review_bot/tasks/base.py
```diff
@@ -7,7 +7,10 @@
 
 logger = structlog.get_logger(__name__)
 
-WORKER_CHECKOUT = '/builds/worker/checkouts/gecko'
+WORKER_CHECKOUTS = (
+    '/builds/worker/checkouts/gecko',
+    '/home/worker/nss',
```
abpostelnicu: We should also add as a checkout path `/home/worker/nspr`

La0:
This is on testing, and just published [a comment on nss-try](https://phabricator-dev.allizom.org/D1277#37577) !

bot/code_review_bot/config.py
```diff
@@ -16,7 +16,8 @@
 import yaml
 
 PROJECT_NAME = 'code-review-bot'
-REPO_TRY = b'https://hg.mozilla.org/try'
+REPO_GECKO_TRY = 'https://hg.mozilla.org/try'
+REPO_NSS_TRY = 'https://hg.mozilla.org/projects/nss-try'
```
marco-c: We need to find a way to remove these when we want to make things more self-service

bot/code_review_bot/revisions.py
```diff
@@ -269,17 +272,25 @@ def setup_try(self, tasks):
         def is_decision_task(task):
             image = task['task']['payload'].get('image')
             if image is not None and isinstance(image, str):
-                return image.startswith('taskcluster/decision')
+                return image.startswith('taskcluster/decision') or \
+                       image.startswith('djmitche/nss-decision')
```
marco-c: Nice image name :P
CC @djmitche

bot/code_review_bot/config.py
```diff
@@ -16,7 +16,8 @@
 import yaml
 
 PROJECT_NAME = 'code-review-bot'
-REPO_TRY = b'https://hg.mozilla.org/try'
```
marco-c: Why was this a bytes? Sure changing it to a string doesn't break anything?

bot/code_review_bot/tasks/base.py
```diff
@@ -92,8 +96,10 @@ def clean_path(self, path):
         '''
         Helper to clean issues path from remote tasks
```
marco-c: I think we should move this stuff to the analyzer themselves, maybe let's file a Bugzilla bug to do so.
Otherwise, we need to list in WORKER_CHECKOUTS all repositories we support.

bot/code_review_bot/config.py
La0: It was bytes to be used by `hglib` to clone from that string. Since i moved to url from secrets, it's no longer used for that purpose . The only usage is now to do the check on decision task.

bot/code_review_bot/tasks/base.py
La0: Completely agreed.

bot/code_review_bot/tasks/coverity.py
```diff
@@ -48,7 +48,7 @@ class CoverityIssue(Issue):
     def __init__(self, revision, issue, file_path):
         self.revision = revision
         self.path = file_path
-        self.reliability = Reliability(issue['reliability'])
+        self.reliability = Reliability(issue['reliability']) if 'reliability' in issue else Reliability.Unknown
```
marco-c: Don't we want the reliability for NSS too? It seems like very useful information!

bot/code_review_bot/config.py
La0: All of that will disappear if we add our task directly into the taskgroup, instead of relying on pulse trigger.
If we can create the review bot's task directly, we can configure its payload to have the build target directly + all relevant infos: no need for finding what task does what

marco-c: :+1:

bot/code_review_bot/tasks/coverity.py
La0: @abpostelnicu ?

La0: That would involve a patch on nss

abpostelnicu:
Will patch nss for that but I don’t like that we have in two places reliability index stored, in one place for m-c and in the second place in the nss repo.

abpostelnicu:
Thinking about reliability a bit more maybe it’s better to have it stored in different places related with the repo that it is used since each project may tune their own reliability indexes depending on the way how coverity can be modeled on the respective projects.
My action item here is to write a patch, it is trivial.

La0:
Let's wait a bit before landing this PR then.

bot/code_review_bot/revisions.py
djmitche: Yeah, this was supposed to get rebuilt by someone with push permission, but I've forgotten who that was.

abpostelnicu:
Regarding the reliability index, I'm not sure that we have the data to display relevant information for NSS, since each code is different, so I think for the moment we should use the implicit value Reliability.Unknown.

bot/code_review_bot/tasks/coverity.py
La0: Agreed with @abpostelnicu to use `Reliability.Unknown` on NSS.

bot/code_review_bot/tasks/base.py
La0: I filed [Bug 1564129](https://bugzilla.mozilla.org/show_bug.cgi?id=1564129) & https://github.com/mozilla/code-review/issues/45 to do that.

PR #43, opened 2019-07-05
Don't fail if the latest coverage task was not correctly indexed

### LINKED ISSUE #40: Do not fail on zero coverage download failure
Currently the [whole task](https://tools.taskcluster.net/groups/EZ0521PaSUGJC8Bn4nI0pQ/tasks/fNuUlYmOSVez7hoBpFb9yA/runs/0/logs/public%2Flogs%2Flive.log) fails !

marco-c:
Fixes #40

### DIFF
--- a/bot/code_review_bot/tasks/base.py
+++ b/bot/code_review_bot/tasks/base.py
@@ -54,8 +54,12 @@ class AnalysisTask(object):
             return
 
         # Load the task & status description
-        task_status = queue_service.status(task_id)
-        task_status['task'] = queue_service.task(task_id)
+        try:
+            task_status = queue_service.status(task_id)
+            task_status['task'] = queue_service.task(task_id)
+        except Exception as e:
+            logger.warn('Task not found', task=task_id, error=str(e))
+            return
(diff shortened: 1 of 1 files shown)


bot/code_review_bot/workflow.py
```diff
@@ -224,9 +224,18 @@ def _in_group(dep_id):
             try:
                 if isinstance(dep, type) and issubclass(dep, AnalysisTask):
                     # Build a class instance from its definition and route
-                    task = dep.build_from_route(self.index_service, self.queue_service)
```
La0: `dep.build_from_route` should never crash: the patch would be way cleaner by catching the exception there and adding a log in the `if` case here. 
I don't like having to check on the ZeroCoverageTask , that's not generic & we could have that kind of failures on other parts

marco-c: Since this is a very weird failure, I wanted to keep raising an exception unless strictly necessary and only log for the already known failure.
I don't mind logging always though, so I'll just do that.

PR #71, opened 2019-08-01
bot: Enforce coding & quotes PEP8 rules

La0:


marco-c:
I would use black with pre-commit, so we have a common style between all projects. What do you think?

La0:
I don't care as long as we are on the same page on every project.
This is just a port of release-services behaviour.

@marco-c Do you have black already setup as you want on bugbug or somewhere else ?

marco-c:
> @marco-c Do you have black already setup as you want on bugbug or somewhere else ?

Yeah, on bugbug, relman-auto-nag and microannotate

marco-c:
E.g. https://github.com/mozilla/microannotate/blob/master/.pre-commit-config.yaml.

PR #73, opened 2019-08-05
Use pre-commit + black + other hooks

La0:


La0:
Because it failed the `Tests should end in _test.py` hook

PR #84, opened 2019-08-21
bot: For the build error reporting phase we strip the `caretline` from the message.

abpostelnicu:
We don't want to have messages like:

where the `caretline` field only creates confusion.


La0:
Demo OK on https://phabricator-dev.allizom.org/harbormaster/unit/1782/

Issue #27, opened 2019-06-11
Support all source-test-*


marco-c:
There are some (very few I guess) we still don't support. For example, https://searchfox.org/mozilla-central/rev/b756e6d00728dda4121f8278a744381d8643317a/taskcluster/ci/source-test/webidl.yml (see it in action at https://treeherder.mozilla.org/#/jobs?repo=mozilla-central&searchStr=webidl&revision=79e6ed0b08d67523705ca61981437874340a38f6&selectedJob=240889848).

The best way to support all would probably be to just have a standard output format, so as soon as a new one is added it would be automatically supported by our bot.

PR #89, opened 2019-08-23
Support all source-test without crashing, refs #27

La0:
Needed to avoid crashes due to node debugger linter in development (already landed the code-review attribute in tree) https://bugzilla.mozilla.org/show_bug.cgi?id=1575295#c3


La0:
self merging now to avoid more crashes

Issue #49, opened 2019-07-17
Enable responses automatically for all tests


marco-c:
We can use pytest-responses to do this.

Adrian-Tamas:
Hi, although I am newish to python, I would like to help with this issue. Can you please detail what needs to be done other than adding pytest-responses in the dependencies and ensure that the unit tests still work or fix failing ones?

La0:
Hello @Adrian-Tamas ,

Adding the `pytest-responses` dep in `bot/requirements-dev.txt`, then removing the `responses.activate` calls should be enough.

To run the tests you'll then need to run `pytest` and check that every test passes.

Adrian-Tamas:
Thanks for the quick reply. I will look into this

PR #91, opened 2019-08-28
Add pytest-responses to enable responses automatically refs #49

Adrian-Tamas:


bot/requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 pre-commit==1.18.2
 pytest==5.1.1
 responses==0.10.6
+pytest-responses==0.4.0
```
La0: We try to keep this file ordered alphabetically: please move that line before `responses`

Adrian-Tamas: fixed

Adrian-Tamas:
@La0 I fixed the order and also removed all the other instances of @responses.activate . Sorry about missing thouse. I did realize that I should have checked in more places but for some reason it was only late yesterday. I have fixed that and squashed the commits into 1.

PR #115, opened 2019-09-24
Papertrail support

La0:
- Move bot's tools in a shared lib, like for code-coverage
- Use our own naming scheme for papertrail
- Use those tools in bot
- Use those tools in events instead of libmozevent logging


bot/Dockerfile
```diff
@@ -1,6 +1,11 @@
 FROM python:3-alpine
 
-COPY bot /src
-RUN pip install /src
+RUN apk add --update gcc musl-dev libffi-dev openssl-dev
```
marco-c: This will make the image quite bigger, you could do:
```
ADD tools /src/tools
ADD bot /src/bot

RUN apk add --update ... && cd /src/tools && python setup.py install && ... && cleanup apk
```

bot/docker/bootstrap.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Add build deps
+apk --update add --virtual build-dependencies gcc musl-dev libffi-dev openssl-dev
+
+# Install python dependencies
+cd /src/tools && python setup.py install
+cd /src/bot && python setup.py install
+
+# Cleanup
+apk del build-dependencies
```
marco-c: Is this going to remove all of `gcc musl-dev libffi-dev openssl-dev`?

marco-c: I'm assuming yes, but I'm not so familiar with `apk`.

La0: [Yeap](https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#virtual-packages) !

bot/Dockerfile
La0: [Before](https://tools.taskcluster.net/groups/BR2ToA1oTO2g3uT74hD_8Q/tasks/BtMg_AiDRHuT1ZbhsgQhcw/runs/0/artifacts): 87,37 Mb
[Now](https://tools.taskcluster.net/groups/Cgj4BMKpSf2qogC7lb1-0w/tasks/eGgJ8uxuSeqRdZMiHhnMDQ/runs/0/artifacts): 41,49 M :tada:

PR #118, opened 2019-09-25
tools: Add release & taskcluster infos in Sentry.

La0:
Now that we control Sentry, we can do cool stuff like:
- using package version to detect if a failure still occurs after a deployment
- adding extra info like a Taskcluster task id.

On this [test Sentry issue](https://sentry.prod.mozaws.net/operations/code-review-bot/issues/6359397) you can view the release `1.0.3` and the taskcluster task id `testBastien`

### DIFF
--- a/tools/code_review_tools/log.py
+++ b/tools/code_review_tools/log.py
@@ -50,14 +50,26 @@ def setup_sentry(name, channel, dsn):
     """
 
     # Detect environment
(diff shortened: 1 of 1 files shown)


marco-c:
It'd be great if we could also group Sentry issues using these. I think there is a way to do that, but I'm not sure.

PR #128, opened 2019-09-30
bot: Log email address that is used when reporting build errors.

abpostelnicu:
As @La0 suggested we should add a log with the email that is used when sending a notification email when a build error occurs.

### DIFF
--- a/bot/code_review_bot/report/mail_builderrors.py
+++ b/bot/code_review_bot/report/mail_builderrors.py
@@ -71,6 +71,8 @@ class BuildErrorsReporter(Reporter):
             logger.info("Unable to find the author for commit.")
             return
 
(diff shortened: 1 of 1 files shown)


bot/code_review_bot/report/mail_builderrors.py
```diff
@@ -71,6 +71,10 @@ def publish(self, issues, revision):
             logger.info("Unable to find the author for commit.")
             return
 
+        logger.info(
+            "Sending build_error email to {}.".format(commit["author"]["email"])
+        )
```
La0: Please use the [structlog features](http://www.structlog.org/en/stable/getting-started.html):

```
logger.info("Send build error email", to=commit["author"]["email"])
```

La0:
I tested locally, works fine (got the email address used)

PR #129, opened 2019-09-30
bot: add dedicated markdown for issues that are build errors.

abpostelnicu:
For build errors when we send targeted emails we must used separate markdowns.

### DIFF
--- a/bot/code_review_bot/tasks/coverity.py
+++ b/bot/code_review_bot/tasks/coverity.py
@@ -31,6 +31,11 @@ ISSUE_MARKDOWN = """
 ```
 """
 
+ERROR_MARKDOWN = """
+**Message**: ```{message}```
+**Location**: {location}
+"""
+
 ISSUE_ELEMENT_IN_STACK = """
 - //{file_path}:{line_number}//:
 -- `{path_type}: {description}`.
@@ -63,6 +68,14 @@ class CoverityIssue(Issue):
         self.message = issue["message"]
 
         self.state_on_server = issue["extra"]["stateOnServer"]
+
+        self.body = None
+        self.nb_lines = 1
+
+        # For build errors we don't embed the stack into the message
+        if self.build_error:
+            return
+
         # If we have `stack` in the `try` result then embed it in the message.
         if "stack" in issue["extra"]:
             self.message += ISSUE_RELATION
@@ -77,9 +90,6 @@ class CoverityIssue(Issue):
                     path_type=event["path_type"],
                     description=event["description"],
                 )
-
-        self.body = None
-        self.nb_lines = 1
 
     def __str__(self):
         return "[{}] {} {}".format(self.kind, self.path, self.line)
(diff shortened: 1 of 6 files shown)


bot/tests/conftest.py
```diff
@@ -50,6 +50,11 @@ def __init__(self, nb):
         def as_markdown(self):
             return "This is the mock issue n°{}".format(self.nb)
 
+        def as_error(self):
```
La0: You are re-implementing the logic from the coverity `as_error` here, you need to use directly a `CoverityIssue` to trigger the real code. 
Otherwise the underlying implementation may change and the unit test won't catch the change

To do that, avoid using `mock_issues`, but instanciate directly a list of `CoverityIssue`

bot/code_review_bot/tasks/coverity.py
```diff
@@ -139,6 +144,13 @@ def as_markdown(self):
             is_clang_error=self.is_clang_error() and "yes" or "no",
         )
 
+    def as_error(self):
+        assert self.build_error
```
La0: Add an assertion message and a docstring on `as_error` for consistency

bot/tests/conftest.py
```diff
@@ -65,6 +66,27 @@ def is_build_error(self):
...
+def mock_coverity_issues():
+    """
+    Build a list of Coverity issues
+    """
+
+    issues = []
+    for i in range(2):
+        issue = {
+            "reliability": "high",
+            "line": i % 2,
```
La0: Why not `"line": i,` directly as you only have 2 issues generated

bot/tests/conftest.py
```diff
@@ -65,6 +66,27 @@ def is_build_error(self):
...
 
 
+@pytest.fixture
+def mock_coverity_issues():
+    """
+    Build a list of Coverity issues
+    """
+
+    issues = []
+    for i in range(2):
```
La0: You can directly do: 

```
return [
  CoverityIssue(0, {...}, "path/to/issue")
  for line in range(2)
]
```
Always use [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) when building a simple structure in a loop

bot/tests/test_reporter_mail.py
```diff
@@ -143,4 +147,4 @@ def _check_email(request):
     conf = {"emails": ["test@mozilla.com"]}
     r = BuildErrorsReporter(conf)
 
-    r.publish(mock_issues, mock_revision)
+    r.publish(mock_coverity_issues, mock_revision)
```
La0: :+1:

La0: Add an assert on the `logger.info` about the email address used using [pytest-structlog](https://pypi.org/project/pytest-structlog/). The library is already used in the unit tests

abpostelnicu: This is so nice, asserting the log!

PR #151, opened 2019-10-11
Use common fields across all Issue implementations

### LINKED ISSUE #149: Use common attributes for all Issue implementation
Currently each `Issue` implementation has some small variations in attributes (char vs column, message not always there, linter vs rule vs bug_type, ...)

To communciate efficiently with the backend (and frontend !) it would be better to have a unified inner representation.

Each sub class will call `Issue.__init__` and fill in the common information there.
The `Issue.as_dict` will be simplified and should be renamed `build_extras_informations()`, set in an extras section. 

The frontend will need to use that new structure (only the detailed view of a revision should be affected)

La0:
Fixes #149


bot/code_review_bot/__init__.py
```diff
@@ -100,12 +125,32 @@ def as_error(self):
...
+            "in_patch": self.revision.contains(self),
+            "is_new": self.is_new,
+            "validates": self.validates(),
+            "publishable": self.is_publishable(),
+            "extras": self.build_extra_informations(),
+        }
+
+    def build_extra_informations(self):
+        """
+        Build the extras information as a dict of JSON serializable values
```
marco-c:
```suggestion
        Build the extra information as a dict of JSON serializable values
```

bot/code_review_bot/__init__.py
```diff
@@ -100,12 +125,32 @@ def as_error(self):
...
+            "level": self.level,
+            "message": self.message,
+            "in_patch": self.revision.contains(self),
+            "is_new": self.is_new,
+            "validates": self.validates(),
+            "publishable": self.is_publishable(),
+            "extras": self.build_extra_informations(),
+        }
+
+    def build_extra_informations(self):
```
marco-c:
```suggestion
    def build_extra_information(self):
```

bot/code_review_bot/tasks/coverity.py
```diff
@@ -53,24 +53,27 @@ class CoverityIssue(Issue):
...
+            message=issue["message"],
+        )
         self.reliability = (
             Reliability(issue["reliability"])
             if "reliability" in issue
             else Reliability.Unknown
         )
-        self.line = issue["line"]
         self.build_error = issue.get("build_error", False)
-        self.bug_type = issue["extra"]["category"]
```
marco-c: We don't need this anymore?

bot/code_review_bot/tasks/infer.py
```diff
@@ -21,10 +21,6 @@
 - **Infer check**: {check}
 - **Publishable **: {publishable}
 - **Is new**: {is_new}
-
-```
-{body}
```
marco-c: No body?

bot/code_review_bot/tasks/infer.py
```diff
@@ -82,54 +81,36 @@ def as_text(self):
...
             name=self.message,
-            code="infer.{}".format(self.bug_type),
-            severity=self.kind,
+            code="infer.{}".format(self.check),
+            severity=self.level,
             path=self.path,
             line=self.line,
             char=self.column,
-            description=self.body,
+            description="Infer detected an issue on that line",
```
marco-c:
```suggestion
            description="Infer detected an issue on this line",
```

bot/code_review_bot/tasks/lint.py
```diff
@@ -98,40 +106,25 @@ def as_markdown(self):
...
-            "nb_lines": self.nb_lines,
-            "rule": self.rule,
-            "message": self.message,
-            "validation": {"disabled_rule": self.is_disabled_rule()},
-            "in_patch": self.revision.contains(self),
-            "is_new": self.is_new,
-            "validates": self.validates(),
-            "publishable": self.is_publishable(),
-        }
+        return {"linter": self.linter}
```
marco-c: Maybe in the mozlint case linter should be the analyzer name?

bot/code_review_bot/__init__.py
```diff
@@ -32,17 +32,42 @@ def __init__(self, code, message):
...
     revision = None
 
+    def __init__(
+        self,
+        analyzer: str,
+        revision,
+        path: str,
+        line: int,
+        nb_lines: int,
+        check: str,
```
marco-c: I'd find a better name for this, maybe "rule"?

bot/code_review_bot/tasks/infer.py
La0: It was always `None`, and only used in the admin mail

bot/code_review_bot/__init__.py
La0: It was rule in one analyzer, and check everywhere else

bot/code_review_bot/tasks/coverity.py
La0: It's the display name of the check, but it was never displayed on phabricator or mail. Only exported on the debug report, and then not displayed by the frontend...

bot/code_review_bot/__init__.py
La0: As discussed, let's keep `check` for now

bot/code_review_bot/__init__.py
La0: I'll merge extra infos & extra identifiers

bot/code_review_bot/tasks/lint.py
La0: definitely

bot/code_review_bot/__init__.py
La0: I removed them, as they are not encessary to build the frontend nor the hash (i left the extra identifiers workflow without any implementation just in case)

PR #152, opened 2019-10-11
Add build hash capability

### LINKED ISSUE #144: Load all issues source files & build issue identifying hash in debug report
To build a hash for 2 phases analysis in a remote backend, we need to load each file mentionned by the issues, using HGMO.
The backend should then be able to build the issues hash instantly, without any remote lookup.

La0:
Fixes #144

First real step towards before/after v2.0 :tada:


bot/tests/conftest.py
```diff
@@ -375,3 +376,25 @@ def mock_taskcluster_config():
     from code_review_bot import taskcluster
 
     taskcluster.options = {"rootUrl": "http://taskcluster.test"}
+
+
+@pytest.fixture
+def mock_hgmo():
+    """
+    Mock HGMO raw-file response to buidl a issue hash
```
marco-c:
```suggestion
    Mock HGMO raw-file response to build a issue hash
```

bot/code_review_bot/revisions.py
```diff
@@ -168,6 +170,32 @@ def analyze_patch(self):
...
         )
 
+    def load_file(self, path):
+        """
+        Load a file content at current revision from remote HGMO
+        """
+        # Check in hgmo cache first
+        cache_path = os.path.join(settings.hgmo_cache, path)
+        if os.path.exists(cache_path):
+            return open(cache_path).read()
```
marco-c: Need to close the file here, or you might leak an handle

bot/code_review_bot/config.py
```diff
@@ -49,6 +49,7 @@ def __init__(self):
         self.taskcluster = None
         self.try_task_id = None
         self.try_group_id = None
+        self.hgmo_cache = tempfile.mkdtemp(suffix="hgmo")
```
marco-c: We are not going to cleanup after ourselves, but this should be fine in our current environment.

bot/tests/test_issues.py
```diff
@@ -65,3 +68,38 @@ def test_cleaned_paths(log, path, cleaned_path):
         assert log.has(
             "Cleaned issue absolute path", path=path, name="test-task", level="warning"
         )
+
+
+def test_build_hash(mock_revision, mock_hgmo):
```
marco-c: Maybe it would be better to have a separate file where we put all the hash-related tests, as I can see we might have a lot of them over time (as we find improvements to the algorithm).

marco-c: We should add a few more tests (IIRC we had them in the past for local before/after?), at least one to check that the hash doesn't change when the indentation changes.

bot/code_review_bot/__init__.py
```diff
@@ -100,6 +107,45 @@ def is_publishable(self):
...
+
+        # Build raw content:
+        # 1. lines affected by patch
+        # 2. without any spaces around each line
+        file_lines = file_content.splitlines()
+        start = self.line - 1  # file_lines start at 0, not 1
+        raw_content = "\n".join(
+            [
+                line.strip()
+                for line in file_lines[start : start + self.nb_lines]  # noqa E203
```
marco-c: Nit: Let's blindly follow the coding style enforced by black, that's why we have black after all ;)

bot/tests/test_issues.py
marco-c: I think it'd also be nice to have a test for each analyzer, so we can clearly see how the hash changes according to the extra data (and maybe we could notice that some extra data should not be part of the hash).

bot/code_review_bot/__init__.py
```diff
@@ -100,6 +107,45 @@ def is_publishable(self):
...
+        """
+        assert self.revision is not None, "Missing revision"
+
+        try:
+            # Load all the lines affected by the issue
+            file_content = self.revision.load_file(self.path)
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code == 404:
+                logger.warning(
+                    "Failed to download a file with an issue", path=self.path
```
marco-c: This should never happen, right? Unless the analyzer is mentioning some build file. Maybe make it an error instead of a warning.

bot/code_review_bot/__init__.py
```diff
@@ -100,6 +107,45 @@ def is_publishable(self):
...
+                )
+                return
```
marco-c: What will the user of the function do in this case? Maybe we could still build the hash, but without the line content?

bot/code_review_bot/__init__.py
La0: This is the coding style from black !
The noqa is to pease flake8 too...

bot/code_review_bot/__init__.py
La0: So it happens with clang-tidy currently quite a lot (Andi made a patch to fix that in https://phabricator.services.mozilla.com/D48702

bot/code_review_bot/__init__.py
La0: Good idea, i'll try that

bot/code_review_bot/__init__.py
marco-c: We need to add `ignore = E203` to the .flake8 file (https://github.com/psf/black#slices).

bot/code_review_bot/config.py
La0: I added a cleanup `atexit`

La0:
[Deployed updated version on testing](https://tools.taskcluster.net/groups/HjA3kLlITz-rmcvEaMb9oQ), in shadow mode.

I'll make another update regarding the unit tests structure

bot/tests/test_issues.py
La0: Did the first part, filed https://github.com/mozilla/code-review/issues/155 for the second

bot/code_review_bot/__init__.py
```diff
@@ -100,6 +107,49 @@ def is_publishable(self):
...
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code == 404:
+                logger.warning(
+                    "Failed to download a file with an issue", path=self.path
+                )
+
+                # We still build the hash with empty content
+                file_content = ""
+            else:
+                # When encoutenring another HTTP error, raise the issue
```
marco-c:
```suggestion
                # When encountering another HTTP error, raise the issue
```

bot/code_review_bot/report/__init__.py
```diff
@@ -29,6 +29,8 @@ def get_reporters(configuration):
             if "reporter" not in conf:
                 raise Exception("Missing reporter declaration")
             name = conf["reporter"]
+            if name != "build_error":
+                continue
```
marco-c: What is this change? Maybe it should be in a separate PR?

La0: Remains from a local test, i'm removing it

PR #161, opened 2019-10-18
Add django base web application

La0:
Add an empty django web application, shipped through CI/CD

PR #162, opened 2019-10-18
Store Revision, Diff and related Issues in the backend

La0:
Based on #161


backend/code_review_backend/issues/management/commands/load_issues.py
```diff
@@ -0,0 +1,181 @@
...
+from code_review_backend.issues.models import Repository
+
+logger = logging.getLogger(__name__)
+
+INDEX_PATH = "project.relman.{environment}.code-review.phabricator.diff"
+
+
+def positive_int(x):
+    """
+    Some analyzer may give line & char positions below 0 to indicate a full file
```
marco-c: Maybe we should fix this in the bot, to have consistency.

backend/code_review_backend/issues/management/commands/load_issues.py
```diff
@@ -0,0 +1,181 @@
...
+            token = data.get("continuationToken")
+            if token is None:
+                break
+
+    def load_local_reports(self):
+        for task_id in os.listdir(self.cache_dir):
+            report = json.load(open(os.path.join(self.cache_dir, task_id)))
+            yield task_id, report
+
+    def build_hierarchy(self, data, task_id):
```
marco-c: The name is a bit confusing, maybe `build_revision_and_diff`?

backend/code_review_backend/issues/models.py
```diff
@@ -0,0 +1,87 @@
...
+    class Meta:
+        abstract = True
+        ordering = ("id",)
+
+
+class Repository(PhabricatorModel):
+
+    slug = models.SlugField(unique=True)
+    url = models.URLField(unique=True)
+    try_url = models.URLField(null=True, blank=True)
```
marco-c: Do we need the `try_url`?

backend/code_review_backend/issues/models.py
```diff
@@ -0,0 +1,87 @@
...
+        return f"D{self.id} - {self.title}"
+
+
+class Diff(PhabricatorModel):
+    revision = models.ForeignKey(
+        Revision, related_name="diffs", on_delete=models.CASCADE
+    )
+
+    review_task_id = models.CharField(max_length=30, unique=True)
+    mercurial = models.CharField(max_length=40)
```
marco-c: Nit: `mercurial_hash`

backend/code_review_backend/issues/models.py
La0: I needed it at first to retrieve the file contents on the backend, as the revision is stored on the try project at that stage.
It's not needed now that the hash is built in the bot...

backend/code_review_backend/issues/management/commands/load_issues.py
La0: You're right, this is cleaner. I'm filing an issue for that (it only happened on the new task from sparky)

backend/code_review_backend/issues/models.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __str__(self):
+        return f"Diff {self.id}"
+
+
+class Issue(models.Model):
+    """An issue detected on a Phabricator patch"""
```
marco-c: Not all issues will be linked to a Phabricator patch, for example the issues we will build to pre-populate the database by running the analyzers on the whole tree. So I think we should split the concepts of issue and Phabricator diff (simply making the link between issue and diff optional).

backend/code_review_backend/issues/management/commands/load_issues.py
La0: https://github.com/mozilla/code-review/issues/168

backend/code_review_backend/issues/models.py
marco-c: (also add a test to test_api for creating an issue with no linked diff)

PR #174, opened 2019-10-21
Use default value for check

### LINKED ISSUE #172: build_hash failure on None mozlint check
https://sentry.prod.mozaws.net/operations/code-review-bot/issues/6465123/

```
TypeError: sequence item 3: expected str instance, NoneType found
(4 additional frame(s) were not displayed)
...
  File "code_review_bot/workflow.py", line 122, in publish
    reporter.publish(issues, revision)
  File "code_review_bot/report/debug.py", line 47, in publish
    "issues": [issue.as_dict() for issue in issues],
  File "code_review_bot/report/debug.py", line 47, in <listcomp>
    "issues": [issue.as_dict() for issue in issues],
  File "code_review_bot/__init__.py", line 199, in as_dict
    "hash": self.build_hash(),
  File "code_review_bot/__init__.py", line 146, in build_hash
    payload = ":".join(

TypeError: sequence item 3: expected str instance, NoneType found
```

La0:
Fixes #172

### DIFF
--- a/bot/code_review_bot/__init__.py
+++ b/bot/code_review_bot/__init__.py
@@ -68,7 +68,7 @@ class Issue(abc.ABC):
         self.path = path
         self.line = line
         self.nb_lines = nb_lines
-        self.check = check or "unknown"
+        self.check = check
 
         # Optional common fields
         self.column = column
--- a/bot/code_review_bot/__init__.py
+++ b/bot/code_review_bot/__init__.py
@@ -68,7 +68,7 @@ class Issue(abc.ABC):
         self.path = path
         self.line = line
         self.nb_lines = nb_lines
-        self.check = check
+        self.check = check or "unknown"
 
         # Optional common fields
         self.column = column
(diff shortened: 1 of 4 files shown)


bot/code_review_bot/__init__.py
```diff
@@ -68,7 +68,7 @@ def __init__(
         self.path = path
         self.line = line
         self.nb_lines = nb_lines
-        self.check = check
+        self.check = check or "unknown"
```
marco-c: This basically happens for analyzers which only have one rule, and so they don't specify it.
Instead of "unknown", maybe it'd be better to do one of:
- Use None;
- Use "default";
- Use the analyzer name itself, since that's basically the rule.

Another thing: should we do it here or for MozLint only? I'm not sure.

marco-c: If we go for a default value, we could put it in the function definition directly.

La0: You're right, it makes more sense to only do it in Mozlint. The issue arises because Mozlint supports many different analyzers & their settings

La0:
I used the analyzer as default Mozlint check

PR #175, opened 2019-10-21
Use positive integers for line & nb_lines

### LINKED ISSUE #168: Use positiver integer (or null) for issue line & nb_lines
The backend needs a positive integer or a null (for full file issue).

This should be easy to do now that the `as_dict` method is used by all issues

La0:
Fixes #168

### DIFF
--- a/bot/code_review_bot/__init__.py
+++ b/bot/code_review_bot/__init__.py
@@ -26,10 +26,13 @@ COVERAGE = "coverage"
 COVERITY = "coverity"
 
 
-def positive_int(x):
+def positive_int(name, x):
     """Helper to get a positive integer or None"""
-    if isinstance(x, int) and x >= 0:
-        return x
+    if isinstance(x, int):
+        if x >= 0:
+            return x
+        else:
+            logger.warning(f"Negative {name} value found, defaults to None", value=x)
     return None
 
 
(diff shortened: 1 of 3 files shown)


bot/code_review_bot/__init__.py
```diff
@@ -132,13 +139,15 @@ def build_hash(self):
...
-            ]
-        )
+        if self.line is None or self.nb_lines is None:
+            # Use full file when line is not specified
+            lines = file_lines
+
+        else:
+            # Use a range of lines
+            start = self.line - 1  # file_lines start at 0, not 1
+            lines = file_lines[start : start + self.nb_lines]  # noqa E203
```
marco-c: Add E203 to .flake8 ignore (https://github.com/psf/black/issues/315#issuecomment-395457972), so we don't have to use `noqa`.

bot/code_review_bot/__init__.py
```diff
@@ -26,6 +26,13 @@
 COVERITY = "coverity"
 
 
+def positive_int(x):
+    """Helper to get a positive integer or None"""
```
marco-c: I'd warn when the line is negative, so we can fix the problem in the analyzers (and make them return None for full file issues).

La0: You smart :cookie:

PR #176, opened 2019-10-21
Add a default task parser

### LINKED ISSUE #153: Make a default task parser
Currently, all reviewbot tasks are parsed based their [suite type](https://github.com/mozilla/code-review/blob/master/bot/code_review_bot/workflow.py#L278-L302). Given that a default analysis format [exists](https://github.com/mozilla/code-review/blob/master/docs/analysis_format.md), we should have a default task parser as well (similar to `MozLintTask`).

Without this, the mozlint test type definitions (in taskcluster) will start to get messy since we have to add non-linting related tasks to it (or the other existing source-tests) to get them to run in reviewbot.

La0:
Fixes #153

@sylvestre if you can take a look at [the updated format in the docs](https://github.com/La0/code-review/blob/165f8203a2bf20d5ed44ad4f0e9387c07eea9051/docs/analysis_format.md)

### DIFF
--- a/bot/code_review_bot/tasks/default.py
+++ b/bot/code_review_bot/tasks/default.py
@@ -2,7 +2,6 @@
 import structlog
 from libmozdata.phabricator import LintResult
 
-from code_review_bot import MOZLINT
 from code_review_bot import Issue
 from code_review_bot import taskcluster
 from code_review_bot.tasks.base import AnalysisTask
@@ -25,7 +24,7 @@ ISSUE_MARKDOWN = """
 
 
 class DefaultIssue(Issue):
-    ANALYZER = MOZLINT
+    ANALYZER = "default"
 
     def validates(self):
         """
--- a/bot/code_review_bot/tasks/default.py
+++ b/bot/code_review_bot/tasks/default.py
@@ -80,6 +80,17 @@ class DefaultTask(AnalysisTask):
         Parse issues from a log file content
         """
         assert isinstance(artifacts, dict)
+
+        def default_check(issue):
+            # Use analyzer name when check is not provided
+            # This happens for analyzers who only have one rule
+            # This logic could become the standard once most analyzers
+            # use that format
+            check = issue.get("check")
+            if check:
+                return check
+            return issue.get("analyzer", self.name)
+
         return [
             DefaultIssue(
                 analyzer=issue.get("analyzer", self.name),
@@ -89,7 +100,7 @@ class DefaultTask(AnalysisTask):
                 column=issue["column"],
                 nb_lines=issue.get("nb_lines", 1),
                 level=issue["level"],
-                check=issue["check"],
+                check=default_check(issue),
                 message=issue["message"],
             )
             for artifact in artifacts.values()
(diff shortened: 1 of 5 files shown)


docs/analysis_format.md
```diff
@@ -2,14 +2,17 @@
...
 
-The [default format](https://github.com/mozilla/code-review/blob/1.0.2/bot/code_review_bot/tasks/lint.py#L170) (loosely based on Mozlint format) has the following fields for each issue:
+The file **must** be available as `public/code-review/issues.json`.
 
-* `path` relative to the repository
-* `column` & `lineno` where the issue is happening in the file
+The [default format](https://github.com/mozilla/code-review/blob/1.0.5/bot/code_review_bot/tasks/default.py#L170) (loosely based on Mozlint format) has the following fields for each issue:
+
+* `path` **relative** to the repository
+* `column` & `line` where the issue is happening in the file. They must be positive integers, or `null` when unknown or for a full file.
```
marco-c:
```suggestion
* `column` & `line` where the issue is happening in the file. They must be positive integers, or `null` when unknown or for an issue linked to a full file.
```

bot/code_review_bot/tasks/default.py
```diff
@@ -0,0 +1,113 @@
...
+        return [
+            DefaultIssue(
+                analyzer=issue.get("analyzer", self.name),
+                revision=revision,
+                path=issue["path"],
+                line=issue["line"],
+                column=issue["column"],
+                nb_lines=issue.get("nb_lines", 1),
+                level=issue["level"],
+                check=issue["check"],
```
marco-c: Maybe we should default the check to the analyzer name like we do for MozLint (I guess many analyzers will be single-rule analyzers).

bot/code_review_bot/workflow.py
```diff
@@ -294,9 +295,15 @@ def build_task(self, task_id, task_status):
...
         elif name == "source-test-infer-infer":
             return InferTask(task_id, task_status)
-        elif name.startswith("source-test-"):
-            logger.error(
-                "Unsupported source-test task: will need a local implementation"
-            )
         else:
-            raise Exception("Unsupported task {}".format(name))
+
+            # If default artifacts available > boom
```
marco-c: We should file an issue (if we don't already have one) to move all analyzers to the common format, so we can cleanup a bit the code.

bot/code_review_bot/tasks/default.py
marco-c: If you do this, don't forget to update the docs too

bot/code_review_bot/workflow.py
La0: It probably would be better to file a Bugzilla meta bug ?

bot/code_review_bot/tasks/default.py
La0: Done

bot/code_review_bot/workflow.py
marco-c: Yep! And an issue in this repo to get rid of the code when that lands in mozilla-central

La0: Filed [Bug 1590669](https://bugzilla.mozilla.org/show_bug.cgi?id=1590669) & #180

PR #196, opened 2019-10-29
Fix linting bug after black update

La0:
:man_facepalming:


La0:
Self merged to be able to validate the other PRs.

PR #200, opened 2019-10-31
frontend: Use directly issues analyzer & check

La0:
Now that the bot outputs always the same payloads for Issues, the frontend does not need to clean them up !

### DIFF
--- a/frontend/src/store.js
+++ b/frontend/src/store.js
@@ -147,31 +147,13 @@ export default new Vuex.Store({
       if (state.stats !== null && report.issues) {
         // Calc stats for this report
         state.stats.checks = report.issues.reduce((stats, issue) => {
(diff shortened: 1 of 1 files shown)


frontend/src/store.js
```diff
@@ -147,31 +147,13 @@ export default new Vuex.Store({
...
-          } else if (issue.analyzer === 'Coverity') {
-            check = issue.kind
-          } else if (issue.analyzer === 'coverage') {
-            check = 'zero coverage'
-          } else {
-            console.warn('Unsupported analyzer', issue.analyzer)
-            return
-          }
-          var key = analyzer + '.' + check
+          var key = issue.analyzer + '.' + issue.check
```
marco-c: Nit: `let` (just to make you happy)

Issue #187, opened 2019-10-24
Migration to backend usage


La0:
- [x] Export phabricator revision full url
- [x] ~Store the try jobs group id~ will use TC Index
- [ ] Add and Update diff state (could even be created by events)
- [x] Store all revisions & diffs processed, even without issues
- [x] Export number of issues per diff in serializer
- [x] Export number of new issues per diff in serializer
- [x] Add filters on diff list: title, repository, has some issues
- [x] Dedicated endpoint for diff creation to be able to list revision detail on `/v1/diff` that becomes RO
- [x] API endpoint to list supported repositories
- [x] API endpoint to get overall checker stats
- [ ] API endpoint to get detailed check usage
- [x] Add CORS support for the backend

La0:
The frontend v2 has been merged in #213, will deploy in testing soon.

La0:
Will ship on next release.

PR #201, opened 2019-10-31
Store analyzers Taskcluster group on backend

La0:
Refs #187


marco-c:
Why do we need it? Can't we use the index instead?

La0:
It's for the frontend to display a link directly to the task group.
How could i access it from the TC Index ?

marco-c:
> It's for the frontend to display a link directly to the task group.
How could i access it from the TC Index ?

From the diff ID you can access the TC indexed task, which is part of the group.
If this PR makes it easier to implement, that's OK.

La0:
I'm closing for now, i'll try to use the index directly from the frontend. That will avoid storing deprecated group ids.

PR #202, opened 2019-10-31
Export phabricator revision url

La0:
refs #187


backend/code_review_backend/issues/models.py
```diff
@@ -47,6 +49,10 @@ class Revision(PhabricatorModel):
     def __str__(self):
         return f"D{self.id} - {self.title}"
 
+    @property
+    def phabricator_url(self):
```
marco-c: Nit: maybe specify `phabricator_revision_url`

La0: My idea was to later add `phabricator_url` on the different model if needed (for Repo, Diff, mayber later Build, ..)

PR #215, opened 2019-11-06
Update .taskcluster.yml for community cluster

La0:


imbstack:
Ah, I hadn't realized that you had done this yet! Thanks for this.

I have a similar diff but also tried to convert the uses of `taskcluster.net`. Does [this](https://github.com/mozilla/code-review/compare/taskcluster-community-migration...imbstack:taskcluster-community-migration) look right? I can make a PR with that or you can just steal my changes if so.

Hopefully the changes I made to permissions today will allow the docker builds to run!

PR #220, opened 2019-11-08
Export creation date

La0:


backend/code_review_backend/issues/tests/test_diff.py
```diff
@@ -40,10 +41,17 @@ def setUp(self):
...
+        # so that all diffs in the test have the same date to be able
+        # to compare the payload easily
+        self.now = datetime.utcnow().isoformat() + "Z"
+        Diff.objects.update(created=self.now)
+
     def test_list_diffs(self):
         """
         Check we can list all diffs with their revision
         """
+        self.maxDiff = None
```
marco-c: You're not using this

La0: It's a [debug trigger for the unit tests](https://docs.python.org/3.4/library/unittest.html#unittest.TestCase.maxDiff)... forgot to remove it

PR #231, opened 2019-11-14
Support full file reporting on Phabricator

### LINKED ISSUE #230: Negative line in issues crash the bot
As seen [here](https://firefoxci.taskcluster-artifacts.net/DdmbgzRTSIaqXW_jUdIGkQ/0/public/logs/live_backing.log)

La0:
Fixes #230

### DIFF
--- a/bot/code_review_bot/report/base.py
+++ b/bot/code_review_bot/report/base.py
@@ -9,6 +9,7 @@ import urllib.parse
 from code_review_bot.tasks.clang_format import ClangFormatIssue
 from code_review_bot.tasks.clang_tidy import ClangTidyIssue
 from code_review_bot.tasks.coverity import CoverityIssue
+from code_review_bot.tasks.default import DefaultIssue
 from code_review_bot.tasks.infer import InferIssue
 from code_review_bot.tasks.lint import MozLintIssue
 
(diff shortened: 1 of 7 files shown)


bot/code_review_bot/report/base.py
```diff
@@ -30,6 +31,7 @@
         "defect": " - {nb} found by mozlint",
         "analyzer": " - `./mach lint --warnings path/to/file` (JS/Python/etc)",
     },
+    DefaultIssue: {"defect": " - {nb} found by a generic analyzer"},
```
marco-c: Could we use the name of the analyzer here?

bot/code_review_bot/report/phabricator.py
```diff
@@ -191,7 +191,9 @@ def comment_inline(self, revision, issue, existing_comments=[]):
         comment = {
             "diffID": revision.diff_id,
             "filePath": issue.path,
-            "lineNumber": issue.line,
+            "lineNumber": issue.line
+            if issue.line is not None
+            else 1,  # support full file
```
marco-c: Did you try this out on a sample Phabricator diff? Could you share the link? (I'm curious to see what it looks like, does it expand the whole file to show the comment?)

La0: I tried with 0 [here](https://phabricator.services.mozilla.com/D52765#1611466) but Phabricator did not display the inline comment (still acepting it). So i'm using 1, so it should show up on the first line

bot/code_review_bot/report/base.py
La0: In a followup yes (under #233), as it involves modifying the stats calculation and how we refer to analyzers.

bot/code_review_bot/report/phabricator.py
La0: Here is a sample output with current code: https://phabricator-dev.allizom.org/D1535#inline-20255

La0:
And a test on a file not expanded: https://phabricator-dev.allizom.org/D1534#inline-20267

PR #238, opened 2019-11-15
Support full file issues reported at line 0

### LINKED ISSUE #230: Negative line in issues crash the bot

La0:
With this code, the licence checks are now reported: [Example on phab-dev](https://phabricator-dev.allizom.org/D1541#41757)

The coverage issue are now also considered full file issues (which makes sense too)

Fixes #230


marco-c:
We could instead make the license analyzer follow our JSON standard and make it return None instead of 0 as the line

marco-c:
> We could instead make the license analyzer follow our JSON standard and make it return None instead of 0 as the line

As discussed on IRC, let's support 0 as None, but let's log an error on Sentry so we can fix analyzers that use 0 to mean full file to use None instead, and analyzers who use 0 as first line to add +1 to their lines.

La0:
@marco-c I updated the bot's code to send a warning on line 0, and falling back to full file mode

PR #239, opened 2019-11-15
Support missing bugzilla id in phabricator data

La0:
Phabricator does not always have that information (missing properties)

### DIFF
--- a/bot/tests/test_revisions.py
+++ b/bot/tests/test_revisions.py
@@ -133,3 +133,22 @@ index 83db48f8..84275f99 100644
     assert mock_revision.contains(issue_in_existing_file_added_line)
     assert not mock_revision.contains(issue_in_not_changed_file)
     assert mock_revision.contains(issue_full_file)
+
+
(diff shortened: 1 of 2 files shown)


bot/tests/test_revisions.py
```diff
@@ -133,3 +133,22 @@ def as_phabricator_lint():
...
+    """Test the bugzilla id parsing from phabricator data"""
+
+    # Default value from mock
+    assert mock_revision.bugzilla_id == 1234567
+
+    # Update phabricator data on revision
+    mock_revision.revision["fields"]["bugzilla.bug-id"] = "456789"
+    assert mock_revision.bugzilla_id == 456789
+
+    # On bad data fallback gracefully
```
marco-c: Do you have a real example of this?

La0: Bad data ? No, but it could happen if a client less smart than moz-phab publishes on phabricator a revision: it's a string there

PR #248, opened 2019-11-18
Add a warning when a failed task without issues is detected

### LINKED ISSUE #243: Warn when a task fails but we get no issue from it
I've noticed https://github.com/mozilla/code-review/issues/237 (and https://bugzilla.mozilla.org/show_bug.cgi?id=1596456) manually out of luck (as I was looking at treeherder and noticed a backout because of linting), but that kind of situation should be clearly surfaced.

La0:
1. sends a warning on Sentry
2. add a message on phabricator top comment

Fixes #243

### DIFF
--- a/bot/code_review_bot/tasks/coverage.py
+++ b/bot/code_review_bot/tasks/coverage.py
@@ -86,6 +86,9 @@ class ZeroCoverageTask(AnalysisTask):
     route = "project.releng.services.project.production.code_coverage_bot.latest"
     artifacts = ["public/zero_coverage_report.json"]
 
+    # Do not send any warning when this task does not find any issues
+    is_allowed_empty = True
+
     def parse_issues(self, artifacts, revision):
         zero_coverage_files = {
             file_info["name"]
--- a/bot/code_review_bot/tasks/coverage.py
+++ b/bot/code_review_bot/tasks/coverage.py
@@ -86,9 +86,6 @@ class ZeroCoverageTask(AnalysisTask):
     route = "project.releng.services.project.production.code_coverage_bot.latest"
     artifacts = ["public/zero_coverage_report.json"]
 
-    # Do not send any warning when this task does not find any issues
-    is_allowed_empty = True
-
     def parse_issues(self, artifacts, revision):
         zero_coverage_files = {
             file_info["name"]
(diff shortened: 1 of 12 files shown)


sylvestre:
Would have this catch the infer issue?

Le lun. 18 nov. 2019 à 16:41, Bastien Abadie <notifications@github.com> a
écrit :

marco-c:
> Would have this catch the infer issue?

Yep, I had filed #243 mainly for that (and for the failing license check which was not detected too).

bot/code_review_bot/tasks/coverage.py
```diff
@@ -86,6 +86,9 @@ class ZeroCoverageTask(AnalysisTask):
     route = "project.releng.services.project.production.code_coverage_bot.latest"
     artifacts = ["public/zero_coverage_report.json"]
 
+    # Do not send any warning when this task does not find any issues
+    is_allowed_empty = True
```
marco-c: Do we actually need this? The task here should always be green, so we shouldn't hit the condition anyway, no?

bot/code_review_bot/workflow.py
```diff
@@ -275,8 +276,23 @@ def _in_group(dep_id):
                     stats.report_task(task, task_issues)
                     issues += task_issues
 
-                    for patch in task.build_patches(artifacts):
+                    task_patches = task.build_patches(artifacts)
+                    for patch in task_patches:
                         revision.add_improvement_patch(task.name, patch)
+
+                    # Report a problem when tasks in erroneous state are found
+                    # but no issue or patch has been processed by the bot
```
marco-c: You're not actually checking the state here, are we checking it somewhere else?

La0: I'm not checking it because we are already processing an erroneous task at this stage of the workflow.

bot/code_review_bot/tasks/coverage.py
La0: We need this to keep a generic workflow: the zero code coverage task must be processed when it's green, but that's the only task that can be processed by the bot and not report any issue. Most other tasks are not processed as their state indicate that they did not find any issue

bot/code_review_bot/workflow.py
marco-c: Fixed in 6d7a6e5.

bot/code_review_bot/tasks/coverage.py
marco-c: Fixed in 6d7a6e5.

PR #249, opened 2019-11-18
events: Listen for task completion pulse messages

marco-c:


events/code_review_events/workflow.py
```diff
@@ -257,6 +262,16 @@ def __init__(self, cache_root):
         )
         self.monitoring.register(self.bus)
 
+        # Create pulse listener for unit test failures
+        self.pulse = PulseListener(
```
La0: Please rename to `pulse_unit_tests` or anything with a suffix, as i'll need `pulse_autoland` pretty soon !

events/code_review_events/__init__.py
```diff
@@ -6,3 +6,4 @@
 QUEUE_MONITORING = "monitoring"
 QUEUE_PHABRICATOR_RESULTS = "results"
 QUEUE_WEB_BUILDS = "builds"
+QUEUE_PULSE = "pulse"
```
La0: Please rename to `QUEUE_PULSE_UNIT_TESTS`

events/code_review_events/workflow.py
marco-c: You might be able to use the same listener, depending on how you want to implement that.

marco-c:
Given that we don't know yet how the autoland stuff will be implemented (maybe it'll just look for the same pulse message), I'll merge this as-is and we can change the naming when the time comes.

PR #258, opened 2019-11-22
Update zero code coverage route.

La0:


La0:
Depends on #259

La0:
It's now :green_heart:

PR #260, opened 2019-11-22
Support user black list

### LINKED ISSUE #250: Don't analyze wptsync patches
They already submit them to try to make sure that everything is OK, so it's not worth for us to analyze them again (especially since they are huge stacks and we submit one push per patch in the stack!).

La0:
Fixes #250

### DIFF
--- a/events/code_review_events/workflow.py
+++ b/events/code_review_events/workflow.py
@@ -71,15 +71,15 @@ class CodeReview(PhabricatorActions):
 
         # Load the blacklisted users
         if user_blacklist:
-            self.user_blacklist = self.api.search_users(
-                constraints={"usernames": user_blacklist}
-            )
-            logger.info(
-                "Blacklisted users",
-                names=[u["fields"]["username"] for u in self.user_blacklist],
-            )
+            self.user_blacklist = {
+                user["phid"]: user["fields"]["username"]
+                for user in self.api.search_users(
+                    constraints={"usernames": user_blacklist}
+                )
+            }
+            logger.info("Blacklisted users", names=self.user_blacklist.values())
         else:
-            self.user_blacklist = []
+            self.user_blacklist = {}
             logger.info("No blacklisted user")
 
     def register(self, bus):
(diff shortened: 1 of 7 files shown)


events/code_review_events/workflow.py
```diff
@@ -134,6 +153,23 @@ def get_repositories(self, repositories, cache_root):
                 # Requeue when nothing changed for now
                 await self.bus.send(QUEUE_WEB_BUILDS, build)
 
+    def is_blacklisted(self, revision: dict):
+        """Check if the revision author is in blacklisted"""
+        author_phid = revision["fields"]["authorPHID"]
+        author = next(
```
marco-c: It'd be nicer to make self.user_blacklist a set of PHIDs, so you can simplify this

events/code_review_events/workflow.py
```diff
@@ -247,14 +283,14 @@ def should_run_test_selection(self, build):
         try:
             if self.should_run_test_selection(build):
                 task = self.community_hooks.triggerHook(
-                    "project-relman", "bugbug-test-select", {"DIFF_ID": build.diff_id},
+                    "project-relman", "bugbug-test-select", {"DIFF_ID": build.diff_id}
```
marco-c: I wonder why this was changed now by pre-commit :S

events/requirements.txt
```diff
@@ -1,2 +1,5 @@
 -e ../tools #egg=code-review-tools
 libmozevent==1.0.5
+
+# Use updated libmozdata with search_users
+https://github.com/mozilla/libmozdata/archive/9923f1a6e8aa20a6f042f8808a380eb0f5c87b49.zip #egg=libmozdata
```
marco-c: We should just release a new version

La0: Done: https://pypi.org/project/libmozdata/0.1.63/

events/code_review_events/workflow.py
La0: :bomb:

events/code_review_events/workflow.py
La0: Yes but i'll lose the naming

PR #262, opened 2019-11-25
events: Don't register for pulse notifications when publication is disabled

marco-c:
Since we don't consume messages when publication is disabled: https://github.com/mozilla/code-review/blob/02d31adcbb878d50fb7f8846ae6d29150178bea0/events/code_review_events/workflow.py#L356-L362.


marco-c:
> You break the self.pulse.run() later call when not using the publication

Oops, I forgot about that one...

Issue #211, opened 2019-11-05
Run analyzers on mozilla-central pushes


La0:
We need the analyzers to run on new mozilla-central pushes in order to get the issues present in the repository, and compare that with new patches

Refs #210

La0:
Work notes:

- [Autoland Pulse messages](https://firefox-ci-tc.services.mozilla.com/pulse-messages?bindings%5B0%5D%5Bexchange%5D=exchange%2Ftaskcluster-queue%2Fv1%2Ftask-completed&bindings%5B0%5D%5Bpattern%5D=route.index.gecko.v2.autoland.latest.taskgraph.decision) on firefox-ci frontend

La0:
Current plan is:

1. events listen to generic task group completion: `exchange/taskcluster-queue/v1/task-group-resolved` with `#` yields simple payloads with `taskGroupId`.
2. events retrieve the task definition using the `taskGroupId`, if it's an autoland decision task (`GECKO_HEAD_REPOSITORY=https://hg.mozilla.org/integration/autoland`), trigger a code review bot task with `AUTOLAND_GROUP_id=taskGroupId`
3. the code review bot uses existing logic to parse the tasks in the group, and report the revision as "clean" when the tasks have no issues, otherwise report all issues found as currently.
4. the "clean" information is used to detect if issues are new or not.

La0:
Steps 1, 2 & 3 are done and running in production, [we can see autoland revisions on the frontend](https://code-review.moz.tools/#/?repository=autoland)

Only thing left is using that info in the backend/bot decision

marco-c:
We also need https://github.com/mozilla/code-review/issues/211#issuecomment-549899075, e.g. clang-tidy would need to run on autoland too (ignoring failures, so Tier 3).

La0:
Right, because now we basically have 0 issues on autoland detected

marco-c:
Oh, and also we need to support full runs to pre-populate the DB.

PR #271, opened 2019-11-26
Cleanup revision creation

La0:
- Removes dependency on `PhabricatorAPI` in the constructor
- Explicitly create a revision from a try task + phabricator api

Preparation work for #211 to be able to create `Revision` from an autoland decision task payload

### DIFF
--- a/bot/code_review_bot/revisions.py
+++ b/bot/code_review_bot/revisions.py
@@ -72,21 +72,49 @@ class Revision(object):
     A Phabricator revision to analyze and report on
     """
 
-    def __init__(self, api, try_task):
-        assert isinstance(api, PhabricatorAPI)
-        assert isinstance(try_task, dict)
-        self.repository = None  # a try repo where the revision is stored
-        self.target_repository = None  # the target repo where the patch may land
+    def __init__(
+        self,
+        id,
+        phid=None,
+        diff_id=None,
+        diff_phid=None,
+        revision=None,
+        diff=None,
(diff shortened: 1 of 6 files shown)


bot/code_review_bot/revisions.py
```diff
@@ -73,22 +72,49 @@ class Revision(object):
...
+        self.phid = phid
+        self.diff_id = diff_id
+        self.diff_phid = diff_phid
+        self.build_target_phid = build_target_phid
+        self.mercurial_revision = mercurial_revision
+        self.revision = revision
+        self.diff = diff
+        self.url = url
+
+        # a try repo where the revision is stored
```
marco-c:
```suggestion
        # a repo where the revision is stored
```

bot/code_review_bot/revisions.py
```diff
@@ -105,45 +131,58 @@ def __repr__(self):
...
-        return "https://{}/D{}".format(self.api.hostname, self.id)
-
-    def load_phabricator(self, try_task):
+    @staticmethod
+    def from_try(try_task, phabricator):
         """
         Load identifiers from Phabricator, using the remote task description
         """
+        assert isinstance(phabricator, PhabricatorAPI)
+        assert isinstance(try_task, dict)
```
marco-c: Nit: you could use types in the function definition

La0: Done

PR #272, opened 2019-11-26
Ingest autoland revisions

La0:
Refs #211 

Will need bot & backend modifications.


bot/code_review_bot/revisions.py
```diff
@@ -180,6 +184,67 @@ def from_try(try_task: dict, phabricator: PhabricatorAPI):
...
+                for d in diffs
+                if d["attachments"]["commits"]["commits"][0]["identifier"]
+                == mercurial_revision
+            ),
+            None,
+        )
+        assert (
+            diff is not None
+        ), f"No Phabricator diff found for D{revision_id} and mercurial revision {mercurial_revision}"
+        logger.info("Found phabricator diff", id=diff["id"])
```
marco-c: Why do we need the Phabricator revision/diff in the autoland case? Is it just to avoid larger changes in the bot?

bot/code_review_bot/workflow.py
```diff
@@ -103,6 +102,47 @@ def run(self, revision):
...
+            title=revision.title,
+            mercurial_revision=revision.mercurial_revision,
+        )
+
+        assert (
+            self.backend_api.enabled
+        ), "Backend storage is disabled, no autoland ingestion possible"
+
+        # Find potential issues in the autoland task group
+        # There should be no issues, so the build is OK
```
marco-c: There can be issues in some tasks, e.g. clang-tidy.

bot/code_review_bot/workflow.py
```diff
@@ -103,6 +102,47 @@ def run(self, revision):
...
+        tasks = self.queue_service.listTaskGroup(settings.autoland_group_id)
```
marco-c: Don't you need the continuationToken dance in order to get all tasks?

bot/code_review_bot/workflow.py
```diff
@@ -103,6 +102,47 @@ def run(self, revision):
...
+                task_issues = task.parse_issues(artifacts, revision)
+                logger.info(
+                    "Found {} issues".format(len(task_issues)),
+                    task=task.name,
+                    id=task.id,
+                )
+                issues += task_issues
+
+        # Store the revision & diff in the backend
+        # TODO: publish state (issues or not issues)
```
marco-c: Isn't this goint to be available from the backend?

bot/code_review_bot/workflow.py
```diff
@@ -325,6 +369,8 @@ def build_task(self, task_id, task_status):
             return InferTask(task_id, task_status)
         elif name.startswith("source-test-"):
             logger.error(f"Unsupported {name} task: will need a local implementation")
+        elif skip_unknown is True:
+            logger.info("Skipping unknown task", id=task_id, name=name)
         else:
             raise Exception("Unsupported task {}".format(name))
```
marco-c: Maybe we should always skip unknown tasks, but log with INFO level when we're on autoland, WARNING level when we're on try.

On try there could still be unknown tasks, e.g. if a developer adds jobs to the reviewbot's try push.

bot/code_review_bot/workflow.py
```diff
@@ -202,12 +242,12 @@ def index(self, revision, **kwargs):
...
             )
 
-    def find_issues(self, revision):
+    def find_issues(self, revision, group_id):
         """
         Find all issues on remote Taskcluster task group
         """
         # Load all tasks in task group
-        tasks = self.queue_service.listTaskGroup(settings.try_group_id)
+        tasks = self.queue_service.listTaskGroup(group_id)
```
marco-c: Here too, the continuationToken stuff. Probably in the try case you didn't need it as there were always just few tasks.

bot/code_review_bot/revisions.py
La0: We need it to create the `Diff` on the backend: it's the `Diff` that stores the mercurial revision + repository (hence indicating the Phabricator revision has reached autoland)

bot/code_review_bot/workflow.py
La0: Yeap:
- no issues on autoland revision :heavy_check_mark: clean before state
- some issues on autoland revision :negative_squared_cross_mark: we have to compare the after issues with these ones

bot/code_review_bot/workflow.py
La0: they are now published

bot/code_review_bot/workflow.py
La0: I now manage the log level according to the environment used.

bot/code_review_bot/workflow.py
La0: I now use the official pagination support

bot/code_review_bot/workflow.py
La0: Yeah, it works OK for now

marco-c: It works OK with try, but it won't work with autoland.

marco-c: Ah, you're only using it for try!

marco-c: (It's still worth fixing btw, can you file a follow-up?)

PR #281, opened 2019-11-28
Do not say when 0 issues found on task failures

### LINKED ISSUE #279: Do not say when 0 issues were found
The Phabricator comment should not start with : 
> Code analysis found 0 defects in the diff XXX:

It confuses people a lot, especially because it happens for a task failure (which is bad enough already.)

La0:
Fixes #279

### DIFF
--- a/bot/code_review_bot/report/base.py
+++ b/bot/code_review_bot/report/base.py
@@ -34,8 +34,7 @@ COMMENT_PARTS = {
     DefaultIssue: {"defect": " - {nb} found by a generic analyzer"},
 }
 COMMENT_FAILURE = """
-Code analysis found {defects_total} in the diff {diff_id}{extras_comments}:
-{defects}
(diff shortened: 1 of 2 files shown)


bot/code_review_bot/report/base.py
```diff
@@ -152,14 +151,17 @@ def pluralize(word, nb):
...
+        if nb > 0:
+            comment = COMMENT_FAILURE.format(
+                defects_total=pluralize("defect", nb), diff_id=revision.diff_id
+            )
+        else:
+            comment = ""
+
+        # Add defects
+        if defects:
+            comment += "\n".join(defects) + "\n"
```
marco-c: Before returning the comment at the end of the function, could you assert that it is not empty as a safe-guard?

La0: Done

PR #286, opened 2019-11-29
events: Add autoland trigger

La0:


events/code_review_events/workflow.py
```diff
@@ -219,7 +225,44 @@ def publish_results(self, payload):
         return True
 
     def parse_pulse(self, payload):
-        assert self.publish is True, "Publication disabled"
+
+        routing = payload["routing"]
+
+        # Process autoland payloads
+        if routing["exchange"] == PULSE_TASK_GROUP_RESOLVED:
```
marco-c: It'd be nicer to have this directly in libmozevent, so we don't have to implement the routing ourselves here. Can you file a follow-up?

events/code_review_events/workflow.py
```diff
@@ -167,7 +170,10 @@ def is_blacklisted(self, revision: dict):
         return True
 
     def publish_results(self, payload):
-        assert self.publish is True, "Publication disabled"
+        if not self.publish:
```
marco-c: Nit: we could just skip the `self.api.update_build_target` calls, so even on testing we run all the code except the upload.

events/code_review_events/workflow.py
```diff
@@ -219,7 +225,44 @@ def publish_results(self, payload):
...
+        # Process autoland payloads
+        if routing["exchange"] == PULSE_TASK_GROUP_RESOLVED:
+            try:
+                self.trigger_autoland(payload["body"])
+            except Exception as e:
+                logger.warn(
+                    "Autoland trigger failure", key=routing["key"], error=str(e)
+                )
+        else:
+            logger.info("Skipping pulse message", key=routing["key"])
```
marco-c: This will flood the log with messages, as we have a lot of task completed messages.

events/code_review_events/workflow.py
```diff
@@ -219,7 +225,44 @@ def publish_results(self, payload):
...
+    def trigger_autoland(self, body: dict):
+        """
+        Trigger a code review autoland ingestion task
+        If the task is an autoland decision task
+        """
+        # Load first task in task group, check if it's an autoland
+        queue = taskcluster_config.get_service("queue")
+        task_group_id = body["taskGroupId"]
+        logger.info("Checking autoland task", task_group_id=task_group_id)
+        task = queue.task(task_group_id)
```
marco-c: Could we check directly from the pulse message data rather than making a request to load the task?

events/code_review_events/workflow.py
```diff
@@ -219,7 +225,44 @@ def publish_results(self, payload):
...
 
     def parse_pulse(self, payload):
-        assert self.publish is True, "Publication disabled"
+
+        routing = payload["routing"]
+
+        # Process autoland payloads
+        if routing["exchange"] == PULSE_TASK_GROUP_RESOLVED:
+            try:
+                self.trigger_autoland(payload["body"])
```
marco-c: It'd be good to have an option to disable it (at least at first, then re-enable it once tried and tested)

events/code_review_events/workflow.py
La0: Unfortunately no, the task group payload has just the group id, no routes, nothing (which makes sense as a group is just an id used by many tasks)

marco-c: (The info is likely at least in "other_routes")

events/code_review_events/workflow.py
La0: Done

events/code_review_events/workflow.py
La0: https://github.com/mozilla/libmozevent/issues/28

events/code_review_events/workflow.py
```diff
@@ -314,25 +357,31 @@ def __init__(self, cache_root):
...
             self.webserver.register(self.bus)
 
-            # Create pulse listener for unit test failures
+            # Create pulse listener
+            exchanges = []
+            if taskcluster_config.secrets["autoland_enabled"]:
+                logger.info("Autoland ingestion is enabled")
+                exchanges += [
+                    # autoland ingestion
+                    (PULSE_TASK_GROUP_RESOLVED, "#")
```
marco-c: You can restrict to level 3 task groups, by doing `#.gecko-level-3.#`, so at least you'll only have autoland and mozilla-central

events/code_review_events/workflow.py
La0: It should be implemented in libmozdata : https://github.com/mozilla/libmozdata/issues/175

events/code_review_events/workflow.py
La0: Awesome

events/code_review_events/workflow.py
marco-c: That's a pity :(

events/code_review_events/workflow.py
```diff
@@ -314,25 +357,31 @@ def __init__(self, cache_root):
...
                     taskcluster_config.secrets["pulse_user"],
                     taskcluster_config.secrets["pulse_password"],
                 )
 
                 # Manually register to set queue as redis
                 self.pulse.bus = self.bus
-                self.bus.add_queue(QUEUE_PULSE, redis=True)
             else:
                 self.pulse = None
+            self.bus.add_queue(QUEUE_PULSE, redis=True)
```
marco-c: Don't we only need this when we set pulse, as it was before?

La0:
Depends on https://github.com/mozilla/code-review/pull/287

marco-c:
Could you also file a follow-up to add a test for trigger_autoland? (to be done after we land this and my PR)

events/code_review_events/workflow.py
```diff
@@ -219,7 +225,44 @@ def publish_results(self, payload):
...
+        """
+        Trigger a code review autoland ingestion task
+        If the task is an autoland decision task
+        """
+        # Load first task in task group, check if it's an autoland
+        queue = taskcluster_config.get_service("queue")
+        task_group_id = body["taskGroupId"]
+        logger.info("Checking autoland task", task_group_id=task_group_id)
+        task = queue.task(task_group_id)
+        if "index.gecko.v2.autoland.latest.firefox.decision" not in task["routes"]:
```
marco-c: Nit: Checking GECKO_HEAD_REPOSITORY from task's payload env could be stabler

PR #294, opened 2019-12-02
Use repository urls instead of assuming slugs

La0:
So we don't have to maintain lists of slugs, and can use any repository (even Github).

### DIFF
--- a/bot/code_review_bot/config.py
+++ b/bot/code_review_bot/config.py
@@ -16,9 +16,6 @@ import tempfile
 import structlog
 
 PROJECT_NAME = "code-review-bot"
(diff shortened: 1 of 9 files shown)


bot/code_review_bot/config.py
```diff
@@ -16,9 +16,6 @@
 import structlog
 
 PROJECT_NAME = "code-review-bot"
-REPO_GECKO_TRY = "https://hg.mozilla.org/try"
-REPO_NSS_TRY = "https://hg.mozilla.org/projects/nss-try"
-RAW_FILE_URL = "https://hg.mozilla.org/mozilla-central/raw-file/tip/{}"
```
La0: Note: this was simply not used in the code anymore.

PR #315, opened 2019-12-04
events: Add missing yaml requirement

marco-c:


marco-c:
We didn't catch this with linting because we lint all projects together, so we install this dependency from bot.

PR #319, opened 2019-12-04
Support repository urls

### LINKED ISSUE #318: Support repository urls in frontend
The bot payload now gives the repo's url instead of a slug.

La0:
fixes #318


La0:
Next version will not use that at all :smiley:

PR #321, opened 2019-12-04
Fix clang tidy bad ../../x paths

La0:
[Sample failure](https://firefox-ci-tc.services.mozilla.com/tasks/KAHwfCmqT1asHJij4uCPsA/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FKAHwfCmqT1asHJij4uCPsA%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log)

### DIFF
--- a/bot/code_review_bot/backend.py
+++ b/bot/code_review_bot/backend.py
@@ -85,8 +85,15 @@ class BackendAPI(object):
         assert revision.issues_url is not None, "Missing issues_url on revision"
         published = 0
         for issue in issues:
+            payload = issue.as_dict()
+            if payload["hash"] is None:
+                logger.warning(
+                    "Missing issue hash, cannot publish on backend", issue=str(issue)
+                )
+                continue
+
             try:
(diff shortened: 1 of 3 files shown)


bot/code_review_bot/backend.py
```diff
@@ -83,10 +83,32 @@ def publish_issues(self, issues, revision):
             return
 
         assert revision.issues_url is not None, "Missing issues_url on revision"
+        published = 0
         for issue in issues:
-            issue.on_backend = self.create(revision.issues_url, issue.as_dict())
+            payload = issue.as_dict()
```
marco-c: Nit: can we store the hash as an attribute of issue so we don't have to convert the issue to dict?

La0: When would we build the hash ?

PR #332, opened 2019-12-05
Use EphemeralStorage class from libmozevent to store builds and pushes related to test selection

### LINKED ISSUE #327: Store diff_to_push and task_group_to_build in durable storage
We still want to keep them in memory for performance reasons (when acting on pulse messages, we want to be extra quick).
So we could:
1) Store them in memory and in durable storage;
2) At startup, read them from durable storage to memory.

marco-c:
Fixes #327

Requires a new version of libmozevent, which I will release tomorrow.

### DIFF
--- a/events/code_review_events/bugbug_utils.py
+++ b/events/code_review_events/bugbug_utils.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import asyncio
 import random
 
 import jsone
@@ -7,6 +8,7 @@ import structlog
 from libmozdata.phabricator import UnitResultState
 from libmozevent.phabricator import PhabricatorBuild
 from libmozevent.phabricator import PhabricatorBuildState
+from libmozevent.storage import EphemeralStorage
 
 from code_review_events import QUEUE_BUGBUG
 from code_review_events import QUEUE_BUGBUG_TRY_PUSH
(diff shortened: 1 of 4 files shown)


marco-c:
I've rebased on top of the libmozevent 1.1.0 update, it should be ready now.

events/code_review_events/bugbug_utils.py
```diff
@@ -63,23 +70,34 @@ def __init__(self):
...
 
         # Store the push revision and build, so we can use it after bugbug
         # selects tests to add.
-        self.diff_to_push[int(build.diff_id)] = {
-            "revision": extras["revision"],
-            "treeherder_url": extras["treeherder_url"],
-            "build": build,
-        }
+        await self.diff_to_push.set(
+            int(build.diff_id),
```
marco-c: I've just realized I might need to use strings as keys everywhere, let me test...

marco-c: Yes, I needed to do that, and I've done it now.

events/code_review_events/bugbug_utils.py
```diff
@@ -63,23 +69,34 @@ def __init__(self):
         self.hooks_service = taskcluster_config.get_service("hooks")
         self.queue_service = taskcluster_config.get_service("queue")
 
+    async def setup(self):
+        self.task_group_to_build = await EphemeralStorage.create(
```
La0: :+1:

events/code_review_events/bugbug_utils.py
```diff
@@ -255,7 +272,11 @@ def add_new_jobs(self, revision, selected_tasks):
             return
 
         # If this diff does not belong to a revision we pushed to try, return.
-        if diff_id not in self.diff_to_push:
+        try:
+            push = self.diff_to_push.get(diff_id)
```
La0: NO await on get ?

marco-c: `get` only uses the memory cache for perf reasons, without accessing Redis

PR #334, opened 2019-12-06
Use task names directly in all report comments

### LINKED ISSUE #233: Use analyzer names instead of classes names in Phabricator top comment
As we are moving away from specific classes for each analyzer toward a unified format, we should use the Taskcluster task name (used as analyzer name) in the top Phabricator comment.

This will modify how we group issues (by analyzer name instead of class)

La0:
Fixes #233

### DIFF
--- a/bot/code_review_bot/report/base.py
+++ b/bot/code_review_bot/report/base.py
@@ -85,10 +85,9 @@ class Reporter(object):
         * count their publishable number
         """
 
-        def by_analyzer(i):
-            return i.analyzer
-
-        groups = itertools.groupby(sorted(issues, key=by_analyzer), by_analyzer)
+        groups = itertools.groupby(
+            sorted(issues, key=lambda i: i.analyzer), lambda i: i.analyzer
+        )
 
         def stats(analyzer, items):
             _items = list(items)
@@ -98,8 +97,14 @@ class Reporter(object):
             if _help is None:
                 for regex, help_value in HELP_COMMANDS.items():
                     if isinstance(regex, Pattern) and regex.match(analyzer):
+                        assert (
+                            _help is None
+                        ), f"Duplicate help command found for {analyzer}"
                         _help = help_value
-                        break
+
+            # Strip source-test- to get cleaner names
+            if analyzer.startswith("source-test-"):
+                analyzer = analyzer[12:]
 
             return {
                 "analyzer": analyzer,
--- a/bot/code_review_bot/report/base.py
+++ b/bot/code_review_bot/report/base.py
@@ -4,34 +4,17 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import itertools
+import re
 import urllib.parse
+from typing import Pattern
 
-from code_review_bot.tasks.clang_format import ClangFormatIssue
-from code_review_bot.tasks.clang_tidy import ClangTidyIssue
-from code_review_bot.tasks.coverity import CoverityIssue
-from code_review_bot.tasks.default import DefaultIssue
-from code_review_bot.tasks.infer import InferIssue
-from code_review_bot.tasks.lint import MozLintIssue
-
-COMMENT_PARTS = {
-    ClangTidyIssue: {
-        "defect": " - {nb} found by clang-tidy",
-        "analyzer": " - `./mach static-analysis check {files}` (C/C++)",
-    },
-    InferIssue: {
-        "defect": " - {nb} found by infer",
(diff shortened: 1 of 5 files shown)


bot/code_review_bot/report/base.py
```diff
@@ -4,34 +4,17 @@
...
-        "analyzer": " - `./mach clang-format -s -p {files}` (C/C++)",
-    },
-    MozLintIssue: {
-        "defect": " - {nb} found by mozlint",
-        "analyzer": " - `./mach lint --warnings path/to/file` (JS/Python/etc)",
-    },
-    DefaultIssue: {"defect": " - {nb} found by a generic analyzer"},
+from typing import Pattern
+
+HELP_COMMANDS = {
```
marco-c: Thought for the future: It'd be nice if the task defined the local command to use in its `extra`, so we don't have to hardcode it here.

bot/code_review_bot/report/base.py
```diff
@@ -97,27 +80,38 @@ def requires(self, configuration, *keys):
...
         """
-        groups = itertools.groupby(
-            sorted(issues, key=lambda x: str(x.__class__)), lambda x: x.__class__
-        )
 
-        def stats(items):
+        def by_analyzer(i):
+            return i.analyzer
+
+        groups = itertools.groupby(sorted(issues, key=by_analyzer), by_analyzer)
```
marco-c: Nit: use a lambda

bot/code_review_bot/report/base.py
```diff
@@ -97,27 +80,38 @@ def requires(self, configuration, *keys):
...
+        def stats(analyzer, items):
             _items = list(items)
+
+            # Lookup the help message, supporting regexes
+            _help = HELP_COMMANDS.get(analyzer)
+            if _help is None:
+                for regex, help_value in HELP_COMMANDS.items():
+                    if isinstance(regex, Pattern) and regex.match(analyzer):
+                        _help = help_value
+                        break
```
marco-c: Nit: instead of breaking, we could check all regexes to make sure we have only one that matches.
```
                    if isinstance(regex, Pattern) and regex.match(analyzer):
                        assert _help is None
                        _help = help_value
```

bot/code_review_bot/report/base.py
La0: Totally agree !

bot/tests/test_reporter_phabricator.py
```diff
@@ -316,7 +396,7 @@ def _check_comment_ccov(request):
         )
 
     issue_clang_tidy = ClangTidyIssue(
-        "mock-clang-format",
+        "source-test-clang-tidy",
```
marco-c: Strange change here :P

La0: Someone f*cked up before.

bot/code_review_bot/report/base.py
```diff
@@ -97,27 +80,38 @@ def requires(self, configuration, *keys):
...
+
             return {
+                "analyzer": analyzer,
```
marco-c: We could remove the `source-test-` part when analyzer starts with `source-test-` to have a cleaner name in the comment

marco-c: (here or in `build_comment` directly)

bot/code_review_bot/report/base.py
```diff
@@ -97,27 +80,38 @@ def requires(self, configuration, *keys):
...
                 "publishable": sum([i.is_publishable() for i in _items]),
                 "publishable_paths": list(
                     {i.path for i in _items if i.is_publishable()}
                 ),
             }
 
-        from collections import OrderedDict
-
-        return OrderedDict([(cls, stats(items)) for cls, items in groups])
+        return [stats(analyzer, items) for analyzer, items in groups]
```
marco-c: Super-nit: return an iterator

marco-c: Or even transform the function into a generator, yielding the stats

bot/code_review_bot/report/base.py
La0: https://github.com/mozilla/code-review/issues/339

bot/code_review_bot/report/base.py
La0: At first i made a shared lambda, but :black_large_square: does not allow that.

bot/code_review_bot/report/base.py
La0: Not worth it :smiling_imp:

PR #347, opened 2019-12-12
Add a few more pre-commit checkers

marco-c:


tools/requirements.txt
```diff
@@ -1,12 +1,12 @@
...
-structlog==19.2.0
-taskcluster==24.0.1
-raven==6.10.0
 
 # limit async-timeout version to avoid using the alpha 4.0.0a0
 # still breaking the events build
 async-timeout<4.0,>=3.0
+Logbook==1.5.3
 
 # limits to avoid building with gcc on alpine
```
La0: this comment is not true for the list of requirements below

marco-c: I've removed it.

PR #356, opened 2019-12-16
Display issue checks history graph

### LINKED ISSUE #320: Add endpoint + graph visualization
Should be easy to make a graph like on coverage front to list the number of issues detected per day

La0:
Fixes #320


La0:
Runs on testing (as it needs the backend to be updated too) : https://code-review.stage.mozaws.net/#/stats

backend/code_review_backend/issues/api.py
```diff
@@ -154,6 +156,48 @@ class IssueCheckStats(generics.ListAPIView):
...
+    * globally
+    * per repository
+    * per analyzer
+    * per check
+    """
+
+    serializer_class = HistoryPointSerializer
+
+    # For ease of use, the history is available without pagination
+    # as the SQL request should always fast to calculate
```
marco-c:
```suggestion
    # as the SQL request should be always fast to calculate
```

marco-c:
> Runs on testing (as it needs the backend to be updated too) : https://code-review.stage.mozaws.net/#/stats

Why is the first point in the graph so high?

La0:
Because i imported the existing issues from Taskcluster index on the first day

Issue #364, opened 2019-12-19
Detect never broken analyzers


La0:
Some analyzers on mozilla-central (tier 1) cannot break: they should always be fully green.
So any issue found by that analyzer should be reported.

We need to:
- mark those analyzers in the taskgraph (add something in their `extra` section)
- read that marker in the bot, and post every issue without filtering

La0:
Current plan discussed with Marco on IRC:
1. Mozilla-Central analyzers switch to our format #180 
2. add a field *per issue* to specify how to report that issue, this will also solve the `mozlint --warnings` #32
3. react accordingly in the bot

marco-c:
This is also a dependency of #25.

La0:
Clear winner of #360 backouts list... First priority for next cycle

La0:
The tier level is already availabe in the treeherder tasks extra payload, i'm using that to mark issues as "breaking CI" and should always be reported.

marco-c:
We can start with that, but I'd go with the other option in the long run (marking the importance in the payload) so we can also support #32.

La0:
Next steps would be :

1. Always publish errors on Phabricator - supporting issues outside of the patch as lint 
2. display errors & warnings clearly on the frontend

marco-c:
This is done.

PR #369, opened 2019-12-20
Sanitize issue level to match real world

La0:
Prep work for #364.

---

Use an explicit enum for issue levels, bringing an extra level (on top of error & warning) as **critical**.
The idea is to mark issues as critical when they come from a tier 1 analyzer (in a followup), and always report them.

By bringing that enum, i also noticed some discrepancies in our level usage (we don't _really_ use it actually):

- the clang-tidy parser used a level variable, that was never set (as the level is not set anymore in the current payload, and notes are parsed in mach) : so every clang-tidy is at Warning level.
- the infer payload only has `ERROR` or `WARNING`, not random strings.


backend/code_review_backend/issues/models.py
```diff
@@ -11,7 +11,12 @@
 
 LEVEL_WARNING = "warning"
 LEVEL_ERROR = "error"
-ISSUE_LEVELS = ((LEVEL_WARNING, "Warning"), (LEVEL_ERROR, "Error"))
+LEVEL_CRITICAL = "critical"
+ISSUE_LEVELS = (
+    (LEVEL_WARNING, "Warning"),
+    (LEVEL_ERROR, "Error"),
+    (LEVEL_CRITICAL, "Critical"),
+)
```
marco-c: What if, for simplicity, we keep only two levels: `error` and `warning`? `error` for CI breaking issues, `warning` for all other issues which developers can feel free to ignore.

marco-c: And we'd temporarily force mozlint issues to be errors and all others to be warnings, until we fix it in mozilla-central

La0: They already are errors ! that's why i added that extra level

marco-c: Yep, but the others can be errors too at the moment, instead we could make them warnings and consider "error" as breaking CI and "warning" as not breaking CI. This way we reduce the changes we need to make (both here and in mozilla-central), without losing any feature (as we are currently not using the level basically).

La0: Ok, works for me

La0: Done, i simply removed the backend patch & the critical level (in 5fdfac8)

bot/code_review_bot/__init__.py
```diff
@@ -57,7 +65,7 @@ def __init__(
         check: str,
         column: int = None,
         message: str = None,
-        level: str = "error",
+        level: Level = Level.Error,
```
marco-c: Maybe warning by default, so when we start publishing all error issues we don't risk publishing some we didn't have to publish

bot/code_review_bot/tasks/coverity.py
```diff
@@ -54,7 +55,7 @@ def __init__(self, analyzer, revision, issue, file_path):
             line=issue["line"],
             nb_lines=1,
             check=issue["flag"],
-            level="error",
+            level=Level.Error,
```
marco-c: Coverity should be warnings, at least for now

bot/code_review_bot/tasks/infer.py
```diff
@@ -48,13 +49,10 @@ def __init__(self, analyzer, entry, revision):
             nb_lines=1,
             check=entry["bug_type"],
             column=entry["column"],
-            level=kind,
+            level=Level(kind.lower()),
```
marco-c: Infer too for now should be always warnings

bot/code_review_bot/__init__.py
La0: Done.

bot/code_review_bot/tasks/coverity.py
La0: Done.

bot/code_review_bot/tasks/infer.py
La0: Done.

bot/tests/test_infer.py
```diff
@@ -106,5 +106,7 @@ def test_infer_artifact(version, nb, mock_revision, mock_hgmo):
 
     issues_data = [issue.as_dict() for issue in issues]
 
+    with open(os.path.join(FIXTURES_DIR, f"infer_issues_{version}.json"), "w") as f:
```
marco-c: I guess this is a left over ;)

La0: woops

PR #384, opened 2020-01-03
Load enabled reviewers just once at startup

marco-c:
This way we can avoid loading reviewers from Phabricator for each diff.

### DIFF
--- a/events/code_review_events/workflow.py
+++ b/events/code_review_events/workflow.py
@@ -110,9 +110,6 @@ class CodeReview(PhabricatorActions):
             try:
                 self.load_patches_stack(build)
                 logger.info("Loaded stack of patches", build=str(build))
-
-                self.load_reviewers(build)
-                logger.info("Loaded reviewers", build=str(build))
             except Exception as e:
                 logger.warning(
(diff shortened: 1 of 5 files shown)


events/code_review_events/workflow.py
```diff
@@ -376,7 +373,7 @@ def __init__(self, cache_root):
             else:
                 self.community_monitoring = None
 
-            self.bugbug_utils = BugbugUtils()
+            self.bugbug_utils = BugbugUtils(self.workflow.api)
```
La0: It would be cleaner to load the reviewers in the workflow, so you don't share the Phabricator API to the bugbug utils

marco-c: I did this so that all bugbug-related stuff is encapsulated in BugbugUtils.

marco-c: So if/when we move it to another project (by having another "events", like https://github.com/mozilla/bugbug/pull/1098), it's easier to move.

La0: As discussed on IRC, let's ship it that way until we plan a real move out of Code review infra.
When moving, bugbugutils should extend phabricatorActions like CodeReview class

PR #385, opened 2020-01-03
Trigger risk analysis for specific authors too

### LINKED ISSUE #376: Allow enabling risk analysis for specific authors
For now, we only support enabling it for specific reviewers.

marco-c:
Fixes #376.

Depends on #384.

### DIFF
--- a/events/code_review_events/bugbug_utils.py
+++ b/events/code_review_events/bugbug_utils.py
@@ -125,7 +125,7 @@ class BugbugUtils:
     def should_run_risk_analysis(self, build):
         """
         Check if we should trigger a risk analysis for this revision:
-        * when the revision is being reviewed by one of some specific reviewers
+        * when the reviewers or the author of the revision are in a list of specific users
         """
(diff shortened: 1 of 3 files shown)


events/code_review_events/bugbug_utils.py
```diff
@@ -132,15 +130,14 @@ def should_run_risk_analysis(self, build):
...
+        userPHIDs = [
             reviewer["reviewerPHID"]
             for reviewer in build.revision["attachments"]["reviewers"]["reviewers"]
         ]
 
-        return any(
-            reviewerPHID in self.risk_analysis_reviewers
-            for reviewerPHID in reviewerPHIDs
-        )
+        userPHIDs.append(build.revision["fields"]["authorPHID"])
```
La0: Nit: i would simply add a comment around here saying that we run BugBug for both reviewers & authors

marco-c: I've updated the docstring of the function

PR #388, opened 2020-01-06
Display publishable stats for issues in frontend

### LINKED ISSUE #375: Display in patch & publication state in frontend
https://code-review.moz.tools/#/stats
all Published are = 0

La0:
The backend will store the `in_patch` information, and exports a `publishable` read-only boolean.

For now publishable = in_patch & new_for_revision

This does not affect the publication mechanism in the bot yet (but could easily be switched to that).

Fixes #375 

:warning: I'm working on a migration to import the in_patch state for existing issues


backend/code_review_backend/issues/api.py
```diff
@@ -152,6 +152,9 @@ class IssueCheckStats(generics.ListAPIView):
     queryset = (
         Issue.objects.values("analyzer", "check", "diff__revision__repository__slug")
         .annotate(total=Count("id"))
+        .annotate(
+            publishable=Count("id", filter=Q(in_patch=True) & Q(new_for_revision=True))
```
marco-c: Since we have this info in the bot, maybe we could make the bot upload it?

La0: I think it makes more sense if the backend has all the pieces to decide what's publishable, save that decision, and give it to the bot (which becomes dumb and publish when the backend says so)

marco-c: Agreed, that's what we have to do in the end for before/after. We could move the publishable decision in the backend now, as a first step (so we also avoid this duplicity). We can do it in a separate PR though.

La0: Filed that https://github.com/mozilla/code-review/issues/390

La0:
@marco-c this pr is reviewable, as i'm making an external script that will use a local try clone to get most patches, and update the data through the API.
Using an internal migration depending on HGMO is extremely slow unfortunately

marco-c:
> as i'm making an external script 

Could you add the script to the repo anyway? Or make a gist and link it here?

La0:
I just added the script that uses HGMO + a worker pool, it should take only a few hours in prod.

backend/code_review_backend/issues/api.py
marco-c: Shouldn't it be OR? If the issue is in the patch, it is publishable. If the issue is not in the patch, but it is new, it is publishable.

backend/code_review_backend/issues/serializers.py
```diff
@@ -134,8 +138,6 @@ class IssueCheckSerializer(serializers.Serializer):
     analyzer = serializers.CharField()
     check = serializers.CharField()
     total = serializers.IntegerField()
-
-    # TODO: support publishable stats number once we have hash comparison stored
     publishable = serializers.IntegerField(default=0)
```
marco-c: Nit: rename it publishable_count to differentiate it?

La0: This serializer is only usable on a statistics endpoint, the backend code should be fine. Are you worried about mis-uses in the frontend ?

backend/code_review_backend/issues/api.py
La0: That would make all issues detected in the first patch publishable; to be consistent with current usage, we should only report `in_patch`.
I added the `new_for_revision` to avoid re-publishing exact duplicate in the revision once we switch the bot

La0:
Got some stats on my local dump (ignore the graph)

backend/code_review_backend/issues/api.py
marco-c: Ah, I wrongly remembered new_for_revision was False for all issues in the first diff of a revision.

Anyway, when we will start publishing issues as Phabricator lint issues, Phabricator will not show issues from a previous diff on the current diff, so the `new_for_revision` check will become harmful. It might be harmful now too, as developers might assume an issue doesn't apply anymore if the comment for the current push doesn't contain that issue. Maybe we should just have publishable == in_patch for now.

backend/code_review_backend/issues/serializers.py
marco-c: Yep, worried about misusing because they have the same name. It's really a nit though, feel free to ignore

backend/code_review_backend/issues/api.py
La0: I made `publishable == in_patch` for now.

PR #391, opened 2020-01-06
setup: Publish a Github release using task-boot

La0:
Depends on https://github.com/mozilla/community-tc-config/pull/180

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -43,7 +43,7 @@ tasks:
         then: 'https://api.code-review.testing.moz.tools'
         else: 'https://api.code-review.moz.tools'
 
-    taskboot_image: "mozilla/taskboot:0.1.10"
+    taskboot_image: "mozilla/taskboot:0.2.1"
 
     pip_install: "pip install --disable-pip-version-check --no-cache-dir --quiet"
     python_version: "3.7"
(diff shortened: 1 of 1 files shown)


La0:
I added the moz-tools-bot account & setup the secret, we only need the role setup from above PR

La0:
It works: [test-bastien release published](https://github.com/mozilla/code-review/releases/tag/test-bastien)

.taskcluster.yml
```diff
@@ -527,3 +527,39 @@ tasks:
...
           owner: bastien@mozilla.com
           source: https://github.com/mozilla/code-review
+
+    - $if: 'tasks_for == "github-push" && head_branch[:10] == "refs/tags/"'
+      then:
+        $let:
+          version: {$eval: 'head_branch[10:]'}
+        in:
+          taskId: {$eval: as_slugid("release")}
+          dependencies:
```
marco-c: Maybe we should depend on the test tasks too, so we only release when everything is green

.taskcluster.yml
```diff
@@ -527,3 +527,39 @@ tasks:
...
+            - {$eval: as_slugid("backend_build")}
+            - {$eval: as_slugid("bot_build")}
+            - {$eval: as_slugid("events_build")}
+            - {$eval: as_slugid("frontend_build")}
+            - {$eval: as_slugid("integration_build")}
+          created: {$fromNow: ''}
+          deadline: {$fromNow: '2 hour'}
```
marco-c:
```suggestion
          deadline: {$fromNow: '2 hours'}
```

.taskcluster.yml
La0: The build steps already depend on the tests, so we should be fine

PR #398, opened 2020-01-07
Fix bugs in load_in_patch script on Heroku

La0:
Found 2 bugs when running on Heroku testing:
1. weirdly i get a decode exception using parsepatch there
2. saw a lot of diffs processed multiple times, my sql request lacked a `DISTINCT`

PR #399, opened 2020-01-07
Workaround to get parsepatch to run on alpine

La0:
Followup of #398 

Reported upstream on https://github.com/mozilla/parsepatch/issues/11

### DIFF
--- a/backend/code_review_backend/issues/management/commands/load_in_patch.py
+++ b/backend/code_review_backend/issues/management/commands/load_in_patch.py
@@ -36,7 +36,7 @@ def load_hgmo_patch(diff):
     resp = requests.get(url)
     resp.raise_for_status()
 
-    patch = Patch.parse_patch(resp.content, skip_comments=False)
+    patch = Patch.parse_patch(resp.text, skip_comments=False)
     assert patch != {}, "Empty patch"
     lines = {
         # Use all changes in new files


backend/code_review_backend/issues/management/commands/load_in_patch.py
```diff
@@ -36,7 +36,7 @@ def load_hgmo_patch(diff):
     resp = requests.get(url)
     resp.raise_for_status()
 
-    patch = Patch.parse_patch(resp.content, skip_comments=False)
+    patch = Patch.parse_patch(resp.content.decode("utf-8"), skip_comments=False)
```
marco-c: Might be better to ignore possible decoding errors

marco-c:
```suggestion
    patch = Patch.parse_patch(resp.content.decode("utf-8", "ignore"), skip_comments=False)
```

jcristau: why not `resp.text`?

marco-c: Yep, that works too, it should be equivalent to resp.content.decode("utf-8", "replace").

La0: i tested in alpine, that works nicely, thanks @jcristau !

PR #401, opened 2020-01-07
Issues from treeherder tier 1 are always errors

La0:
Refs #364

Next steps would be : 
1. Always publish errors on Phabricator - supporting issues outside of the patch as lint errors
2. display errors & warnings clearly on the frontend


La0:
Closing because we'll rely on direct levels for now.

PR #403, opened 2020-01-08
events: Handle cases where there are no jobdetails results

marco-c:


events/code_review_events/bugbug_utils.py
```diff
@@ -364,13 +364,16 @@ def add_new_jobs(self, revision, selected_tasks):
...
                 uuid = slugid.decode(status["taskId"])
                 last_run = body["runId"]
                 job_details = self.treeherder_client.get_job_details(
                     job_guid=f"{uuid}/{last_run}"
                 )
-                job_id = job_details[0]["job_id"]
-                treeherder_url = f"https://treeherder.mozilla.org/#/jobs?repo=try&revision={revision}&selectedJob={job_id}"
+                if len(job_details) > 0:
+                    job_id = job_details[0]["job_id"]
+                    treeherder_url += f"&selectedJob={job_id}"
```
La0: Nit: Would be cleaner to use  [urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)

marco-c: Done with 3615784.

PR #410, opened 2020-01-08
Use Python 3.8 for the events and integration Docker images

marco-c:


marco-c:
Depends on #407.

marco-c:
Actually, depends on #420.

PR #417, opened 2020-01-08
Common Phabricator unit & lint results

La0:
Prep work for #364 

Every issue needed to implement `as_phabricator_lint` & `as_phabricator_unitresult`. But we have re-organized the Issue inner data earlier, allowing us to have a unique implementation for these methods.

The unit results are only used by Coverity (for build errors), but every Issue will be able to post a lint result now (for an error outside the patch).

The good news is that the payload do not change a lot (even if they were not used until now) as you can see in the unit tests.

### DIFF
--- a/bot/code_review_bot/tasks/infer.py
+++ b/bot/code_review_bot/tasks/infer.py
@@ -4,7 +4,6 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import structlog
-from libmozdata.phabricator import LintResult
 
 from code_review_bot import Issue
 from code_review_bot import Level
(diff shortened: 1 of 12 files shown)


bot/code_review_bot/tasks/infer.py
```diff
@@ -22,15 +21,6 @@
...
 """
 
-INFER_SETUP_CMD = [
-    "gecko-env",
-    "./mach",
-    "artifact",
-    "toolchain",
-    "--from-build",
-    "linux64-infer",
-]
```
La0: A useless const from an old time :smirk:

Issue #26, opened 2019-04-05
Publish issues that are outside the revision files as generic comments


marco-c:
Since they can't be published as comments linked to a revision file.

La0:
We now use Phabricator unit & lint results to publish errors outside of patch.

Issue #25, opened 2019-03-27
Publish some issues as unignorable lint issues and others as comments


marco-c:
Issues that would break mozilla-central CI can be published as Phabricator lint issues (since they *have to* be fixed in order to land the revision).
Issues coming from SA tools with false positives should be published as comments, so that developers can decide to ignore them.

marco-c:
@smacleod could we prevent landing patches which have unignorable lint issues?

marco-c:
As we discussed recently, we could add a warning to Lando (or even prevent landing completely for unignorable issues, since they will lead to backout if landed).

marco-c:
https://github.com/mozilla/code-review/pull/418 is publishing some errors as lint issues. After we release it to prod and after a few days of usage, we can go all the way and publish all errors as lint issues (see https://github.com/mozilla/code-review/pull/418#discussion_r364714241).

marco-c:
Reopened for https://github.com/mozilla/code-review/issues/25#issuecomment-572570606, to track publishing all errors as lint issues.

La0:
I'm adding a TC secret switch to enable/disable error publication, and will publish all errors as lint results.

PR #418, opened 2020-01-08
Publish errors

La0:
Requires #417 
Refs #364 

This PR will report all errors, even if they are outside a patch (as Phabricator lint results).

:information_source: I'm adding some unit tests to cover that new publication.

### DIFF
--- a/backend/code_review_backend/issues/tests/test_issue.py
+++ b/backend/code_review_backend/issues/tests/test_issue.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from django.test import TestCase
+
+from code_review_backend.issues.models import LEVEL_ERROR
+from code_review_backend.issues.models import LEVEL_WARNING
+from code_review_backend.issues.models import Issue
+
+
+class IssueTestCase(TestCase):
+    def test_publishable(self):
+
+        # A warning is not publishable
+        issue = Issue.objects.create(path="some/file", line=12, level=LEVEL_WARNING)
+        self.assertFalse(issue.publishable)
+
+        # An error is publishable
+        issue = Issue.objects.create(path="some/file", line=12, level=LEVEL_ERROR)
+        self.assertTrue(issue.publishable)
+
+        # A warning in a patch is publishable
+        issue = Issue.objects.create(
+            path="some/file", line=12, level=LEVEL_WARNING, in_patch=True
+        )
+        self.assertTrue(issue.publishable)
--- a/backend/code_review_backend/issues/tests/test_issue.py
+++ b/backend/code_review_backend/issues/tests/test_issue.py
@@ -26,3 +26,9 @@ class IssueTestCase(TestCase):
             path="some/file", line=12, level=LEVEL_WARNING, in_patch=True
         )
         self.assertTrue(issue.publishable)
+
+        # An error in a patch is publishable
+        issue = Issue.objects.create(
+            path="some/file", line=12, level=LEVEL_ERROR, in_patch=True
+        )
+        self.assertTrue(issue.publishable)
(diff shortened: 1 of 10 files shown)


La0:
I've added the unit test to cover error publication

La0:
I created a [test revision](https://phabricator-dev.allizom.org/D1654) to run the bot & publish a lint result from my computer

La0:
And i got a lint result published :tada: (twice as i ran twice on the same build)

https://phabricator-dev.allizom.org/D1654

marco-c:
> And i got a lint result published tada (twice as i ran twice on the same build)

Great!
Could you try updating the diff and seeing if the previous lint issues are cleared?

backend/code_review_backend/issues/tests/test_issue.py
```diff
@@ -0,0 +1,28 @@
...
+
+        # An error is publishable
+        issue = Issue.objects.create(path="some/file", line=12, level=LEVEL_ERROR)
+        self.assertTrue(issue.publishable)
+
+        # A warning in a patch is publishable
+        issue = Issue.objects.create(
+            path="some/file", line=12, level=LEVEL_WARNING, in_patch=True
+        )
+        self.assertTrue(issue.publishable)
```
marco-c: Nit: for completeness, also test an error in a patch is publishable

bot/code_review_bot/report/phabricator.py
```diff
@@ -70,25 +78,39 @@ def publish(self, issues, revision, task_failures):
             # Always publish comment summarizing issues
             self.publish_comment(revision, issues_only, patches, task_failures)
 
-            # Also publish build issues
-            if self.publish_build_errors:
-                self.publish_harbormaster_build_errors(revision, build_errors)
+            # Publish all errors outside of the patch as lint issues
```
marco-c: Maybe we should publish all errors as lint issues, no matter if they are outside or inside the patch?

bot/code_review_bot/report/phabricator.py
```diff
@@ -70,25 +78,39 @@ def publish(self, issues, revision, task_failures):
...
-            logger.info("No build errors encountered")
+        if not lint_issues and not unit_issues:
+            logger.info("No issues to publish as Phabricator lint or unit results")
             return
+
         self.api.update_build_target(
             revision.build_target_phid,
-            BuildState.Fail,
-            unit=[issue.as_phabricator_unitresult() for issue in issues],
+            state=BuildState.Work,
```
marco-c: Why Work and not Fail?

marco-c:
This would fix #25.

marco-c:
And it would fix #26 too.

marco-c:
> And it would fix #26 too.

Actually not, for #26 we need more changes.

bot/code_review_bot/report/phabricator.py
La0: Fail is sent afterward by the workflow. It should not be the responsibility of the phabricator reporter to give the final state (as it's set in stone once fail or pass).

Currently, a revision with a build error gets 2 Fail status, which is weird

bot/code_review_bot/report/phabricator.py
La0: I would start with only extra errors as lint issues, see how the dev react, then upgrade to that later on.

backend/code_review_backend/issues/tests/test_issue.py
La0: Done

bot/code_review_bot/report/phabricator.py
marco-c: > Fail is sent afterward by the workflow

Do you mean https://github.com/mozilla/code-review/blob/d570e38cea78f7479d5d0f6b81934a6689ef169a/bot/code_review_bot/workflow.py#L200? Does this one cover the build failure case?

bot/code_review_bot/report/phabricator.py
marco-c: OK, we can use #25 to track that.

marco-c: https://github.com/mozilla/code-review/issues/25#issuecomment-572570606

La0:
>  Could you try updating the diff and seeing if the previous lint issues are cleared?

I pushed a new diff that use the `fm` variable, the errors automatically cleared from the display.

bot/code_review_bot/report/phabricator.py
La0: Yes, `nb_publishable` will be positive for any in patch warnings or errors anywhere

PR #420, opened 2020-01-08
Freeze library versions in tools

### LINKED ISSUE #414: Freeze library versions in tools/
This way builds are more reproducible.

marco-c:
Fixes #414


tools/requirements.txt
```diff
@@ -1,8 +1,8 @@
-# limit async-timeout version to avoid using the alpha 4.0.0a0
-async-timeout<4.0,>=3.0
-Logbook==1.5.3
-multidict<4.6.0
+aiohttp==3.6.2
```
marco-c: I've moved it here since it's also installed in the bot via the taskcluster dep

PR #433, opened 2020-01-13
events: Support passing the Phabricator deployment to bugbug

marco-c:
Depends on https://github.com/mozilla/bugbug/pull/1243.

### DIFF
--- a/events/code_review_events/bugbug_utils.py
+++ b/events/code_review_events/bugbug_utils.py
@@ -29,6 +29,12 @@ EPHEMERAL_STORAGE_EXPIRATION = 25200
 
 class BugbugUtils:
     def __init__(self, phabricator_api):
+        self.phabricator_deployment = (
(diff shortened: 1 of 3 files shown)


events/code_review_events/bugbug_utils.py
```diff
@@ -29,6 +29,12 @@
 
 class BugbugUtils:
     def __init__(self, phabricator_api):
+        self.phabricator_deployment = (
```
La0: I would prefer using `APP_CHANNEL` here or a dedicated secret instead of checking the phabricator url

marco-c: The problem with APP_CHANNEL is that sometimes we point testing to look at Phabricator prod and we wouldn't support that

marco-c: I've switched to a secret.

PR #434, opened 2020-01-13
Display issues found by a check

### LINKED ISSUE #416: Display issues per check
Restore the feature from the frontend v1 where a user could list all the issues found by a check.

The list will be of course paginated using the API

La0:
Fixes #416


La0:
Current render on a check:

PR #436, opened 2020-01-14
Use a single Redis connection

marco-c:
This is more or less how things would look like with https://github.com/mozilla/libmozevent/pull/55.


marco-c:
This is now no longer needed given https://github.com/mozilla/libmozevent/pull/70.

PR #457, opened 2020-02-04
events: Pass runnable-jobs from the decision task when triggering a bugbug test selection task

marco-c:
This way bugbug can skip jobs which are not runnable.
Note: this will make test selection start a little later, as we start it after pushing to try rather than when we get the notification from Phabricator, but this should be a limited delay.

### DIFF
--- a/events/code_review_events/bugbug_utils.py
+++ b/events/code_review_events/bugbug_utils.py
@@ -119,15 +119,15 @@ class BugbugUtils:
             },
         )
 
+        # Start test selection
+        await self.start_test_selection(build, extras["revision"])
+
     async def process_build(self, build):
         assert build is not None, "Invalid payload"
         assert isinstance(build, PhabricatorBuild)
 
         # Start risk analysis
         await self.start_risk_analysis(build)
-
-        # Start test selection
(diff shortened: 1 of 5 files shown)


events/code_review_events/bugbug_utils.py
```diff
@@ -199,6 +199,7 @@ def should_run_test_selection(self, build):
                 {
                     "PHABRICATOR_DEPLOYMENT": self.phabricator_deployment,
                     "DIFF_ID": build.diff_id,
+                    "RUNNABLE_JOBS": f"https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.try.revision.{revision}.firefox.decision/artifacts/public/runnable-jobs.json",
```
La0: I would prefer to use the configured taskcluster buildUrl function `taskcluster_config.get_service('index').buildUrl('findArtifactFromTask', 'gecko.v2.try.revision.{revision}.firefox.decision', 'public/runnable-jobs.json')`

marco-c: Good idea, done with 29bf899.

I've also added a test.

PR #462, opened 2020-02-07
Support TEST-UNEXPECTED-FAIL from logs

### LINKED ISSUE #41: Support unit test results
In order to support unit tests reporting through the code-review workflow, the bot needs to publish unit tests failures on Phabricator.
This is needed for [Bug 1561292](https://bugzilla.mozilla.org/show_bug.cgi?id=1561292)

Here is a proposal for a JSON artifact payload for unit tests: 

```json
{
  "path/to/test/file.py": [
    {
      "type": "unittest",
      "name": "AdditionTest",
      "result": "fail",
      "engine": "pytest",
      "details": "1+1 should be equal to 2, not 3",
    }
  ],
}
```

The result field should [match Phabricator](https://phabricator-dev.allizom.org/conduit/method/harbormaster.sendmessage/) : 

* pass
* fail
* skip
* broken
* unsound

La0:
Fixes #41 

I'm not sure this is the best way to tackle this as TEST-UNEXPECTED-FAIL usually lacks the file path & test name.

### DIFF
--- a/bot/code_review_bot/__init__.py
+++ b/bot/code_review_bot/__init__.py
@@ -254,15 +254,19 @@ class Issue(abc.ABC):
         """
         Build a Phabricator UnitResult for build errors
         """
-        assert (
-            self.is_build_error()
-        ), "Only build errors may be published as unit results"
+        if self.is_build_error():
+            details = f"Code review bot found a **build error**: \n{self.message}"
+            name = "general"
+        else:
+            details = self.message
+            name = self.check or "test"
 
         return UnitResult(
             namespace="code-review",
-            name="general",
+            name=name,
+            engine=self.analyzer,
             result=UnitResultState.Fail,
-            details=f"Code review bot found a **build error**: \n{self.message}",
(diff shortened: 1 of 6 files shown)


marco-c:
What about doing https://github.com/mozilla/code-review/issues/41#issuecomment-582498793? This way we don't need to parse the logs, and we have a centralized place for that (Treeherder).

La0:
Well i started doing that, but I think it's not generic enough: we won't support unit tests with that, just another way of saying a task failed.

It would be better to add json output to the debugger-tests task itself

marco-c:
> Well i started doing that, but I think it's not generic enough: we won't support unit tests with that, just another way of saying a task failed.

It's actually *more* generic, as it supports any failure that Treeherder supports.

> It would be better to add json output to the debugger-tests task itself

I think it would be a huge task to make all test reporters support the same format, relying on Treeherder seems better here as the work is done for us (and done once in a single place).

PR #463, opened 2020-02-07
Use treeherder url for task failures

### LINKED ISSUE #41: Support unit test results

La0:
Hopefully almost fixes #41 ...

Several things are happening here:
1. i'm sharing the treeherder job url between events & bot - it's setup in code_review_tools
2. events is updated to use that method
3. the bot uses the treeherder url for a task failure (instead of the Taskcluster frontend view) - displaying potential issues
4. A `DefaultTask` is used for all `source-test-*` tasks, meaning that any failing task from `source-test` will use the task failure mechanism.

A followup would be to report such failures through use unit results instead of a comment.

### DIFF
--- a/bot/code_review_bot/report/base.py
+++ b/bot/code_review_bot/report/base.py
@@ -8,6 +8,8 @@ import re
 import urllib.parse
 from typing import Pattern
 
+from code_review_tools import treeherder
+
 HELP_COMMANDS = {
     "source-test-clang-tidy": " - `./mach static-analysis check {files}` (C/C++)",
     "source-test-infer-infer": " - `./mach static-analysis check-java path/to/file.java` (Java)",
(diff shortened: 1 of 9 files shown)


marco-c:
> A followup would be to report such failures through use unit results instead of a comment.

And then we could do https://github.com/mozilla/code-review/issues/313, so we show the failure lines from Treeherder.

PR #470, opened 2020-02-17
tools: Limit idna version to fix docker build.

La0:
Should fix [this issue](https://community-tc.services.mozilla.com/tasks/ZQD3iy3fTQagBaDWelclvg/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FZQD3iy3fTQagBaDWelclvg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log)

### DIFF
--- a/tools/requirements.txt
+++ b/tools/requirements.txt
@@ -1,5 +1,8 @@
 aiohttp==3.6.2
 async-timeout==3.0.1
+
+# Limit idna to avid conflicts
+idna<2.9,>=2.5
 logbook==1.5.3
 multidict==4.5.2
 raven==6.10.0


tools/requirements.txt
```diff
@@ -1,5 +1,8 @@
 aiohttp==3.6.2
 async-timeout==3.0.1
+
+# Limit idna to avid conflicts
```
marco-c:
```suggestion
# Limit idna to avoid conflicts
```

PR #481, opened 2020-02-19
Use task group id to find decision task

### LINKED ISSUE #480: Detect automatically decision task
By finding the top parent, or simply assuming the task using taskGroupId as taskId is the decision task.

Blocks #461

La0:
Fixes #480

In the end, i used the heuristic `decision task id == try group id`, as it's possible to create tasks from the decision tasks without adding the decision task as a dependency, thus making it impossible to trace it back

### DIFF
--- a/bot/code_review_bot/workflow.py
+++ b/bot/code_review_bot/workflow.py
@@ -266,12 +266,10 @@ class Workflow(object):
         assert "tasks" in tasks
         tasks = {task["status"]["taskId"]: task for task in tasks["tasks"]}
         assert len(tasks) > 0
-        logger.info(
-            "Loaded Taskcluster group", id=settings.try_group_id, tasks=len(tasks)
-        )
+        logger.info("Loaded Taskcluster group", id=group_id, tasks=len(tasks))
 
         # Update the local revision with tasks
-        revision.setup_try(tasks)
+        revision.setup_try(group_id, tasks)
 
         # Store the revision in the backend
         # It needs to be after setup_try to have a repository value
--- a/bot/tests/conftest.py
+++ b/bot/tests/conftest.py
@@ -386,7 +386,7 @@ def mock_backend_secret(mock_taskcluster_config):
 
 
 @pytest.fixture
-def mock_workflow(mock_phabricator, mock_taskcluster_config):
+def mock_workflow(mock_config, mock_phabricator, mock_taskcluster_config):
     """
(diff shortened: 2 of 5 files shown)


bot/code_review_bot/workflow.py
```diff
@@ -266,12 +266,10 @@ def find_issues(self, revision, group_id):
...
         tasks = {task["status"]["taskId"]: task for task in tasks["tasks"]}
         assert len(tasks) > 0
-        logger.info(
-            "Loaded Taskcluster group", id=settings.try_group_id, tasks=len(tasks)
-        )
+        logger.info("Loaded Taskcluster group", id=group_id, tasks=len(tasks))
 
         # Update the local revision with tasks
-        revision.setup_try(tasks)
+        revision.setup_try(group_id, tasks)
```
marco-c: A little below, we are already getting the decision task with `task = tasks.get(settings.try_task_id)`. Maybe make `setup_try` return the decision task, so we have no duplicated code (and if we decide to change the way to detect the decision task, we can do it just once).

marco-c: Or, get the decision task here and pass it to setup_try too.

bot/tests/conftest.py
```diff
@@ -414,6 +414,9 @@ def setup_mock_tasks(self, tasks):
             """
             Add mock tasks in queue & index mock services
             """
+            # The task group id is used to find the decision task
+            # as it should be the first task in the group
```
marco-c: The comment here should say "it's the task with the same ID as the group"

bot/code_review_bot/workflow.py
La0: the task designed by `settings.try_task_id` is not the decision task, but the code-review one. It's used to list all the analyzers with potential issues

marco-c: Oh, right.

bot/tests/test_remote.py
```diff
@@ -250,16 +250,16 @@ def test_decision_task(mock_config, mock_revision, mock_workflow, mock_backend):
     """
 
     mock_workflow.setup_mock_tasks({"decision": {}, "remoteTryTask": {}})
```
marco-c: We no longer have any test for the `Missing decision task` case, could you add it?

La0: Done

PR #482, opened 2020-02-19
Use repositories config from secret to find mercurial infos in decision task

### LINKED ISSUE #479: Store decision tasks environment keys in secret
To easily manage new projects, the existing `repositories` payload should have 2 more keys per repository:

1. `decision_key_revision` to find the value for `mercurial_revision` of the bot's revision (example `GECKO_HEAD_REV`)
2. `decision_key_repository` to find the value for `repository` of the bot's repository (example `GECKO_HEAD_REPOSITORY`)
3. we can use the existing `url` key as `target`

Blocks #461

La0:
Fixes #479

When releasing, i'll simply need to move the existing `repositories` configuration in secret in **common**  section, and then add the relevant decision keys.

### DIFF
--- a/bot/code_review_bot/revisions.py
+++ b/bot/code_review_bot/revisions.py
@@ -18,7 +18,6 @@ from code_review_bot import taskcluster
 from code_review_bot.config import REGEX_PHABRICATOR_COMMIT
 from code_review_bot.config import REPO_AUTOLAND
 from code_review_bot.config import REPO_MOZILLA_CENTRAL
-from code_review_bot.config import REPO_NSS
 from code_review_bot.config import settings
 
 logger = structlog.get_logger(__name__)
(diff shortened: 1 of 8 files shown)


bot/code_review_bot/revisions.py
```diff
@@ -406,29 +405,27 @@ def is_decision_task(task):
...
-            # Unfortunately the NSS decision task does not expose the target repository
-            self.target_repository = REPO_NSS
-
-        else:
-            raise Exception("Unsupported decision task")
-
-        # Save mercurial revision
-        assert self.mercurial_revision is not None, "Missing try revision"
-        logger.info("Using Try mercurial revision", rev=self.mercurial_revision)
+        for repository in settings.repositories:
```
marco-c: What if we used the information in the Phabricator revision to detect which repository to use?

La0: It won't work because Phabricator has no knowledge of the try repositories, and i need the try & target repos + mercurial revision.

marco-c: I meant just in order to detect which try repository to use, instead of relying on the presence of the env variable in the decision task.

La0: Done: i'm using the Phabricator repo info to match the decision task

Issue #475, opened 2020-02-18
Build and use bot Docker image directly in firefox-ci


La0:
Instead of pushing the bot's docker image on dockerhub, we need to build it with DinD on the firefox-ci instance, mark that task with a route, and use the testing/production images in the hook.
This will allow us to ship releases without any intervention from releng (see Bug [1594102](https://bugzilla.mozilla.org/show_bug.cgi?id=1594102))

The plan is :
- [x] Enable Firefox-ci Github app on the repo - [Bug 1616209](https://bugzilla.mozilla.org/show_bug.cgi?id=1616209)
- [x] Update `.taskcluster.yml` to build DinD image on firefox-ci instance #476 
- [x] Add route to the DinD build task on firefox-ci instance #488 
- [ ] Update hooks to use the docker images with `index-image`

I'll certainly have to request some extra scopes on firefox-ci, for the differents tasks to run & maybe some personal scopes for admins.

Ping @tomprince to show you the plan

La0:
The firefoxci TC github app is now enabled here.

tomprince:
That plan seems reasonable.

La0:
The [testing hook](https://firefox-ci-tc.services.mozilla.com/hooks/project-relman/code-review-testing) has been updated.

La0:
The [production hook](https://firefox-ci-tc.services.mozilla.com/hooks/project-relman/code-review-production) has been updated, and code review bot is running OK.

PR #486, opened 2020-02-21
Build bot docker image with DinD on firefox-ci Taskcluster instance

La0:
Refs #475


marco-c:
We are running linting and unit tests for the bot on both community and firefox-ci, maybe we should run them only in community

La0:
That's what i did initially, but the docker build depends on that (which makes sense).
After the next release, i'll remove every bot job from the community instance, as we won't need to build the bot docker image there (nor push it anymore)

Issue #487, opened 2020-02-21
Store Phabricator builds from events


La0:
This would be shipped after release 1.1.6.
The idea is to make a request from events on the backend, on every new build being processed, in order to store the Phabricator build & target ids, on the `Diff` model.
It should be pretty straightforward, as we already have access to the phabricator API from events, and are able to get informations for Diff, Revision & Repository.

The requests should be made right after the applied patch has been pushed on the try server (as we would then have the mercurial revision to use - which is required to create a Diff).

The `review_task_id` would become nullable as it's not created at the point in the patch timeline.

(Optional) A `state` field should also be added on the `Diff` model, as an enum:
- `pushed` when the try push has just been made
- `done` when the analysis has been done

It would then be possible to query the backend looking for a Diff on a try server at a specific mercurial revision (which is the only information we are certain to have on the decision task payload that the bot parses).

Would also help for #468, #461, #366

marco-c:
`events` was removed in https://github.com/mozilla/code-review/pull/2652.

Issue #251, opened 2019-11-18
Stop getting both phabricator-diff and phabricator-build-target


marco-c:
We can just decide about one and get rid of the code for the other.

PR #490, opened 2020-02-24
Read existing try_task_config on HGMO to support all repos

La0:
While working on #487, it became evident that we already store the phabricator build id for every repo on try, and it would be quick to use that information.

Several things change with that PR:
1. All the informations needed to initialize the `Revision` instance are fetched at the earliest stage, before running the workflow : no more `setup_try` later on.
2. The bot would rely on HGMO to start & support both variable names (should do #251 at some point)
3. Most of the `Issue` hashes in the unit tests change because the revision & repo are used to generate a mock file, and are set to `None` currently (i can make a PR to fix that before landing this one, it would be cleaner & the hashes should not change then)
4. I have to compare the repository from secret with the decision task environment (checking keys), as i can't know the right repository anymore (from Phabricator)

It's a WIP, i need to run it on several revisions from different repos, and a new unit test file to cover most cases from the `from_try` method.

I think #487 is still useful, but this could be a short term solution to support CI repositories easily


bot/tests/test_try.py
```diff
@@ -0,0 +1,192 @@
...
+    decision_task = {
+        "metadata": {"name": "Fake decision task"},
+        "payload": {"env": {"KEY": "VALUE"}},
+    }
+    with pytest.raises(AssertionError) as e:
+        Revision.from_try(decision_task, None)
+    assert str(e.value) == "Unsupported decision task"
+
+
+def test_mozila_central(mock_config, mock_phabricator, mock_hgmo):
```
marco-c:
```suggestion
def test_mozilla_central(mock_config, mock_phabricator, mock_hgmo):
```

bot/code_review_bot/revisions.py
```diff
@@ -134,20 +152,52 @@ def __str__(self):
...
-    def from_try(try_task: dict, phabricator: PhabricatorAPI):
+    def from_try(try_decision_task: dict, phabricator: PhabricatorAPI):
         """
         Load identifiers from Phabricator, using the remote task description
         """
+        logger.info(
+            "Using try decision task", name=try_decision_task["metadata"]["name"]
+        )
+
+        def match_task_env(repository):
```
marco-c: Could we keep matching the task with the Phabricator repository? This way we avoid issues where two repos have the same envs

La0: I don't really see how: at that point in the timeline, we only have access to the tasks definition (and the mercurial informations buried in it). before we had access to the specific phabricator build id, which gave me the specific repository. Now that's the opposite

marco-c: Ah because we don't have the URL to get try_task_config.json, that's painful.

PR #501, opened 2020-02-25
Use standard indexes.

tomprince:
This depends on https://phabricator.services.mozilla.com/D64152 landing for PRs to work.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -149,12 +149,13 @@ tasks:
                   path: /bot.tar
                   type: file
             routes:
-              - "index.project.code-review.bot.rev.${head_rev}"
-              - $if: 'tasks_for == "github-pull-request"'
-                then: "index.project.code-review.bot.pull_request.${head_branch}"
-                else: "index.project.code-review.bot.${head_branch}"
-            scopes:
-              - queue:route:index.project.code-review.bot.*
+              $if: 'tasks_for == "github-pull-request"'
+              then:
+                - "index.code-analysis.v2.code-review-pr.revision.${head_rev}"
+                - "index.code-analysis.v2.code-review-pr.branch.${head_branch}"
+              else:
+                - "index.code-analysis.v2.code-review.revision.${head_rev}"
+                - "index.code-analysis.v2.code-review.branch.${head_branch}"
             metadata:
               name: Code Review Bot docker in docker build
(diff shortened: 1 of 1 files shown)


La0:
Are those standard index format documented somewhere ? I thought the `project.X.Y.Z` was still valid as [it's still used by many tasks](https://firefox-ci-tc.services.mozilla.com/tasks/index/project)

tomprince:
> Are those standard index format documented somewhere ? I thought the `project.X.Y.Z` was still valid as [it's still used by many tasks](https://firefox-ci-tc.services.mozilla.com/tasks/index/project)

It is still used in some places, but we grant `{trust_domain}.v2.{alias}.*` to repos via the `trust-domain-scopes` features, (along with generic scopes for worker-types and caches). What goes after that prefix isn't well defined[1], and I just picked something similar to what you had under `project.*`. I don't have strong feelings about anything other than using the new prefix that doesn't require additional grants.

[1] taskgraph has some defaults, but doesn't enforce them. https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/transforms/task.py#l189 are the default indexes that taskgraph creates (but note they are based on hg rather than github projects)

.taskcluster.yml
```diff
@@ -146,10 +149,14 @@ tasks:
...
-                then: "index.project.code-review.bot.pull_request.${head_branch}"
-                else: "index.project.code-review.bot.${head_branch}"
+              $if: 'tasks_for == "github-pull-request"'
+              then:
+                - "index.code-analysis.v2.code-review-pr.revision.${head_rev}"
+                - "index.code-analysis.v2.code-review-pr.branch.${head_branch}"
+              else:
+                - "index.code-analysis.v2.code-review.revision.${head_rev}"
+                - "index.code-analysis.v2.code-review.branch.${head_branch}"
+                - "code-analysis.v2.code-review.revsion.${head_rev}"
```
La0:
```suggestion

```

.taskcluster.yml
```diff
@@ -146,10 +149,14 @@ tasks:
...
             scopes:
               - queue:route:index.project.code-review.bot.*
```
La0: This scope needs to be removed or updated

La0:
I can't update the pr on your fork, i may re-create it in order to fix the scope request

PR #503, opened 2020-02-26
Support all dependant tasks by default

### LINKED ISSUE #502: Configure default task trigger from secret, by repository
In order to support tasks without the default artifact, but with task failure detection, we need to be able to filter the tasks by name (we always use `source-test-` prefix for MC and others).

For example, taskgraph would need a filter like `tests-*`.
This filtre should be configurable on the `repositories` secret, and default to `*`

La0:
Fixes #502


La0:
I ran on several tasks on MC, no issue whatsoever, and was able to post on takgraph too.

PR #506, opened 2020-02-26
Fix black setup on CI

La0:
The latest virtualenv release breaks black setup in pre-commit.


La0:
I created an [issue on virtualenv](https://github.com/pypa/virtualenv/issues/1669)

PR #507, opened 2020-02-26
Do not build or ship the bot from Community TC

La0:
- Only test & build on Firefox CI
- No more bot build with img on Community CI
- No more bot push on Community CI

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -86,6 +86,8 @@ tasks:
               owner: bastien@mozilla.com
               source: https://github.com/mozilla/code-review
 
+        # Run only on firefoxci Taskcluster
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -606,7 +537,6 @@ tasks:
                 taskId: {$eval: as_slugid("release")}
                 dependencies:
                   - {$eval: as_slugid("backend_build")}
-                  - {$eval: as_slugid("bot_build")}
```
marco-c: This is a bit of a drawback, but I don't see any solution other than moving everything to Firefox-CI.

La0: That's the long term goal, but that won't happen soon, as the images built with the old DinD are not compatible with Heroku

PR #514, opened 2020-02-27
Get repository and revision from code-review task preferentially.

tomprince:


tomprince:
(missing tests)

La0:
We can revert https://github.com/mozilla/code-review/commit/81053bf1953ced537ce91739057d8f6835b9d86f now that you're implementing the same behaviour as on MC.
@marco-c This way we won't have to maintain env variables. WDYT ?

marco-c:
Sounds good to me. It means all repos we'll want to support will have to implement the same, but it's probably an OK assumption.

PR #539, opened 2020-03-18
bot: Add script to trigger missing tasks

La0:


bot/tools/fix_missing.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Move to next page
+        url = data["next"]
+
+
+def timestamp(date):
+    return time.mktime(date.timetuple())
+
+
+def list_pushes(known_revisions, updates, min_date, max_date):
```
marco-c: You could probably list pushes from hgmo directly, but this is fine too

PR #545, opened 2020-03-23
Use common mockup class for Phabricator API

### LINKED ISSUE #421: Use shared function to check phabricator calls in unit tests
The file `bot/tests/test_reporter_phabricator.py` is becoming pretty big, with a lot of repetition of code to check the Phabricator mocks behaviour; we should be able to refactor that.

La0:
Fixes #421 

Pave the way to use lint/unit results instead of comments.


La0:
I know it's almost impossible to review properly :cry: That certainly says a lot about the state of that test file !

Issue #339, opened 2019-12-09
Support help command from source-test tasks


La0:
Every analyzer should add an `extra` key in its task payload in order to explain how to reproduce the analysis locally : the bot should support this and remove local `HELP_COMMANDS`

La0:
It should also support a `display_name` extra variable. Both would be used as a primary source for the AnalysisTasks.

PR #555, opened 2020-03-25
Use nicer analyzer names on Phabricator

### LINKED ISSUE #554: Use nicer analyzer names
When publishing stats in the summary comment:
  - `source-test-` is removed
  - `coverage` becomes `code coverage analysis`

When publishing a Phabricator lint result, the `source-test-mozlint-analyzer` analyzer should be `analyzer (from mozlint)`.

These rules should be implemented in the `name` method (or a new `display_name`) at the `AnalysisTask` level. Said tasks should then be propagated through the issues, instead of simply passing the analyzer key (or be accessible in another way).

La0:
Fixes #554

This PR allow us to customize the analyzer name displayed in Phabricator (and other parts of the system).
The `AnalysisTask` instances are now fully accessible from their issues (instead of a single string), which allows us to have some logic from the tasks accessible when building payloads for Phabricator.

Two new methods are available on the `AnalysisTask` sub classes:
- `build_help_message` customizes the optional help message in the summary comment
- `display_name` is a nicer name, only used for humans (:robot: still use the shorter and unique `name`).

### DIFF
--- a/bot/code_review_bot/tasks/base.py
+++ b/bot/code_review_bot/tasks/base.py
@@ -31,11 +31,27 @@ class AnalysisTask(object):
 
     @property
     def name(self):
+        """Short name used to identify the task"""
         return self.task["metadata"].get("name", "unknown")
+
+    @property
+    def display_name(self):
+        """
+        Longer name used to describe the task to humans
+        By default fallback to short name
+        """
+        return self.name
 
     @property
     def state(self):
         return self.status["state"]
+
+    def build_help_message(self, files):
+        """
+        An optional help message aimed at developers to reproduce the issues detection
+        A list of relative paths with issues is specified to build a precise message
+        By default it's empty (None)
+        """
 
     @classmethod
     def build_from_route(cls, index_service, queue_service):
--- a/bot/code_review_bot/tasks/clang_tidy.py
+++ b/bot/code_review_bot/tasks/clang_tidy.py
@@ -155,10 +155,18 @@ class ClangTidyTask(AnalysisTask):
 
     artifacts = ["public/code-review/clang-tidy.json"]
 
+    @property
+    def display_name(self):
+        return "clang-tidy"
+
+    def build_help_message(self, files):
+        files = " ".join(files)
+        return f"`./mach static-analysis check {files}` (C/C++)"
+
     def parse_issues(self, artifacts, revision):
         return [
             ClangTidyIssue(
-                analyzer=self.name,
+                analyzer=self,
                 revision=revision,
                 path=path,
                 line=warning["line"],
--- a/bot/code_review_bot/tasks/lint.py
+++ b/bot/code_review_bot/tasks/lint.py
@@ -34,12 +34,12 @@ class MozLintIssue(Issue):
         message,
         check,
         revision,
-        **kwargs
+        **kwargs,
     ):
         # Use analyzer name when check is not provided
         # This happens for analyzers who only have one rule
         if check is None:
-            check = analyzer
+            check = analyzer.name
         super().__init__(
             analyzer,
             revision,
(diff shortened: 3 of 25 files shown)


bot/code_review_bot/tasks/clang_tidy.py
```diff
@@ -155,10 +155,18 @@ class ClangTidyTask(AnalysisTask):
 
     artifacts = ["public/code-review/clang-tidy.json"]
 
+    @property
+    def display_name(self):
+        return "clang-tidy"
+
+    def build_help_message(self, files):
+        files = " ".join(files)
+        return f"`./mach static-analysis check {files}` (C/C++)"
```
sylvestre: I have been thinking about that. No need to do that here 
but I would be interested to tell the developer if the lint has an autofix or not
for example, mozlint/flake8 with --fix will call autopep8

bot/code_review_bot/tasks/lint.py
```diff
@@ -110,14 +110,32 @@ class MozLintTask(AnalysisTask):
...
+    def linter(self):
+        # Detect linter from task name
+        if not self.name.startswith("source-test-mozlint-"):
+            return
+        return self.name[20:]
+
+    @property
+    def display_name(self):
+        if self.linter:
+            return f"{self.linter} (by MozLint)"
```
marco-c: Nit:
```suggestion
            return f"{self.linter} (Mozlint)"
```

bot/code_review_bot/tasks/lint.py
```diff
@@ -110,14 +110,32 @@ class MozLintTask(AnalysisTask):
...
+        if self.name.startswith("source-test-"):
+            return self.name[12:]
+        return self.name
+
+    def build_help_message(self, files):
+        return "`./mach lint --warnings --outgoing` (JS/Python/etc)"
```
marco-c:
```suggestion
        return "`./mach lint --warnings --outgoing`"
```

bot/code_review_bot/tasks/base.py
```diff
@@ -31,12 +31,28 @@ def run_id(self):
...
     @property
     def state(self):
         return self.status["state"]
 
+    def build_help_message(self, files):
+        """
+        An optional help message aimed at developers to reproduce the issues detection
+        A list of relative paths with issues is specified to build a precise message
+        By default it's empty (None)
+        """
```
marco-c: We could make display_name and help_message optional values of the standard JSON format that checkers have to provide.
Otherwise, we'll never be able to be fully generic and get rid of the custom AnalysisTask classes.

bot/code_review_bot/tasks/lint.py
sylvestre:
```suggestion
            return f"{self.linter} (by Mozlint)"
```
https://firefox-source-docs.mozilla.org/code-quality/lint/index.html

bot/tests/test_reporter_phabricator.py
```diff
@@ -49,7 +55,7 @@
 
 VALID_FLAKE8_MESSAGE = """
 Code analysis found 1 defect in the diff 42:
- - 1 defect found by mozlint-py-flake8
+ - 1 defect found by py-flake8 (by MozLint)
```
sylvestre:
```suggestion
 - 1 defect found by py-flake8 (by Mozlint)
```

bot/tests/test_reporter_phabricator.py
```diff
@@ -94,7 +100,7 @@
 
 VALID_MOZLINT_MESSAGE = """
 Code analysis found 2 defects in the diff 42:
- - 2 defects found by mozlint-dummy
+ - 2 defects found by dummy (by MozLint)
```
sylvestre:
```suggestion
 - 2 defects found by dummy (by Mozlint)
```

bot/tests/test_reporter_phabricator.py
```diff
@@ -729,7 +776,7 @@ def test_extra_errors(mock_phabricator, mock_try_task, phab):
                     "code": "EYYY",
                     "description": "Some not so bad python mistake",
                     "line": 2,
-                    "name": "source-test-mozlint-dummy",
+                    "name": "dummy (by MozLint)",
```
sylvestre:
```suggestion
                    "name": "dummy (by Mozlint)",
```

bot/code_review_bot/tasks/clang_tidy.py
La0: Only sane way to do that is if the help message comes from the task itself: https://github.com/mozilla/code-review/issues/339

bot/code_review_bot/tasks/base.py
La0: :arrow_right: #339 

I'll do that as a followup, with documentation.

PR #558, opened 2020-03-27
bot: for build errors push them as lint issues.

### LINKED ISSUE #552: Publish build errors as lint results
When we can parse a build error and point to the file where the build error is, we could use a lint issue instead of a unit test result.

abpostelnicu:
For the `unitresult` we only push the `message` from `CoverityIssue`, in order to make it more verbose to be able to better understand what the actual issue is we can have a format like:

`[file path]:[line number] message`

Fixes #552.


bot/tests/test_backend.py
```diff
@@ -82,10 +82,10 @@ def test_publication(mock_coverity_issues, mock_revision, mock_backend, mock_hgm
             "analyzer": "mock-coverity",
             "check": "flag",
             "column": None,
-            "hash": "1fcc4d02d6184028f40b48e877be62b4",
+            "hash": "172f015fbc43268d712c2fc7acbf1023",
             "id": "98d7e3b0-e903-57e3-9973-d11d3a9849f4",
             "in_patch": False,
-            "level": "warning",
+            "level": "error",
```
marco-c: Why did these two turn into errors? Were they build errors?

bot/tests/test_remote.py
```diff
@@ -696,21 +694,15 @@ def test_coverity_task(
...
-        format="remarkup",
-    )
+
     assert check_stats(
         [
             ("code-review.analysis.files", None, 2),
             ("code-review.analysis.lines", None, 2),
             ("code-review.issues", "source-test-coverity-coverity", 2),
-            ("code-review.issues.publishable", "source-test-coverity-coverity", 0),
+            ("code-review.issues.publishable", "source-test-coverity-coverity", 1),
```
marco-c: I'd say the build errors should be in a different statistics than normal coverity errors

bot/code_review_bot/report/phabricator.py
```diff
@@ -67,13 +67,12 @@ def publish(self, issues, revision, task_failures):
         if issues or task_failures:
             # Split issues and build errors
             issues_only = [issue for issue in issues if not issue.is_build_error()]
```
La0: You can remove that and use issues everywhere

abpostelnicu: sure

bot/tests/test_remote.py
abpostelnicu: Why do you say this, since in the end they are published.

marco-c: It would be useful to be able to differentiate between statistics about Coverity issues and about build errors.

La0: That can be in a followup i think.

bot/tests/test_reporter_phabricator.py
```diff
@@ -579,6 +579,10 @@ def test_phabricator_unitresult(mock_phabricator, phab, mock_try_task, mock_task
             "test.cpp": [41, 42, 43]
         }
         revision.build_target_phid = "PHID-HMBD-deadbeef12456"
+        revision.repository = "https://hg.mozilla.org/try"
```
La0: Please update the test name to `test_phabricator_coverity`

bot/tests/test_reporter_phabricator.py
```diff
@@ -579,6 +579,10 @@ def test_phabricator_unitresult(mock_phabricator, phab, mock_try_task, mock_task
...
+        revision.repository_try_name = "try"
+        revision.mercurial_revision = "deadbeef1234"
+
         reporter = PhabricatorReporter(
             {"analyzers": ["coverity"], "publish_build_errors": True}, api=api
```
La0: The config can now be just `{}`

bot/tests/test_reporter_phabricator.py
```diff
@@ -627,17 +631,19 @@ def test_phabricator_unitresult(mock_phabricator, phab, mock_try_task, mock_task
         assert phab.build_messages["PHID-HMBD-deadbeef12456"] == [
```
La0: We need to check the summary comment here too. We have no idea how the coverity issues are listed now

bot/tests/test_remote.py
marco-c: We should do it soon though, or our stats will get biased

Issue #553, opened 2020-03-25
Add a comment when errors are published outside of patch


La0:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1610715#c6

When we publish errors outside of patch, developers can be confused why they are reported.
We should add an extra info in the summary comment: ` Code analysis found X defects in the diff Y (defects might be in the parent stack)` in such a case

marco-c:
It'd be nice to avoid reporting issues that are in a parent patch, as they are already reported in the parent patch. I wonder if we could list issues from the parent and skip them. Or maybe somehow do a try push which considers only the latest patch.

junngo:
I submitted PR :)
Please let me know if there is anything that needs to be modified.

bhushan-borole:
Hello, I would like to work on this issue.

marco-c:
@junngo submitted a PR to fix this a while ago, https://github.com/mozilla/code-review/pull/561, though it seems it's been stuck for a while.
@junngo do you still want to continue or do you want @bhushan-borole to take over?

junngo:
Hello @marco-c 
I think @bhushan-borole can take this issue.

bhushan-borole:
Thats great @junngo.
I will start working on it @marco-c

PR #561, opened 2020-03-30
Add a comment when errors are published outside of patch

junngo:
#553 
Added a comment when errors are published outside of patch.

**before**
Code analysis found X defects in the diff Y

**after**
Code analysis found X defects in the diff Y **(defects might be in the parent stack)**

### DIFF
--- a/bot/code_review_bot/report/base.py
+++ b/bot/code_review_bot/report/base.py
@@ -146,6 +146,11 @@ class Reporter(object):
         else:
             comment = ""
 
+        if not all(
+            revision.contains(issue) for issue in issues if issue.is_publishable()
+        ):
+            comment += "(defects might be in the parent stack)\n"
+
         # Add defects
         if defects:
             comment += "\n".join(defects) + "\n"
--- a/bot/tests/test_reporter_phabricator.py
+++ b/bot/tests/test_reporter_phabricator.py
@@ -121,6 +121,7 @@ If you see a problem in this automated review, [please report it here](https://b
 
 VALID_MOZLINT_MESSAGE = """
 Code analysis found 2 defects in the diff 42:
+(defects might be in the parent stack)
  - 2 defects found by dummy (Mozlint)
 
 You can run this analysis locally with:


La0:
Thanks for your PR, it's a great start !
This issue was filed to add the extra string only when errors are pubilshed outside of patch. So you need to detect that, instead of always adding the string.

You can know if the current revision contains an issue with `revision.contains(issues)`

junngo:
Thanks for reviewing :)
Is it be okay to stick it on when the issue is included in the revision?
A case that is a little hard for me to understand is the case when errors are published outside of patch.
May I ask you a question? I wonder if there are cases where errors are not published as a rough example.

bot/code_review_bot/report/base.py
```diff
@@ -146,6 +146,9 @@ def pluralize(word, nb):
         else:
             comment = ""
 
+        if any(revision.contains(issue) for issue in issues):
```
La0, line 149: That's actually the opposite test: you need to add the sentence, when the revision does not contain every publishable issue:

```python
if not all(revision.contains(issue) for issue in issues if issue.is_publishable):
 ...
```

bot/code_review_bot/report/base.py
```diff
@@ -146,6 +146,9 @@ def pluralize(word, nb):
...
+            comment = comment.replace(":", " (defects might be in the parent stack):")
```
La0, line 150: You should just add the content to comment:

```python
comment += " (defects might be in the parent stack
```

bot/tests/test_reporter_phabricator.py
```diff
@@ -27,7 +27,7 @@
 from code_review_bot.tasks.lint import MozLintTask
 
 VALID_CLANG_TIDY_MESSAGE = """
-Code analysis found 1 defect in the diff 42:
+Code analysis found 1 defect in the diff 42 (defects might be in the parent stack):
```
La0, line 30: These cases seem to be in the patch, so they should **not** have the sentence

bot/code_review_bot/report/base.py
junngo: I want to modify it like that, but a new line(`\n`) occurs.
iss_533_newline
If you modify it like a review, it goes down a line.
Is that okay?

junngo:
Thanks for the good review.
I submitted PR again. My understanding has become clearer thanks for reviewing.

bot/code_review_bot/report/base.py
```diff
@@ -146,6 +146,9 @@ def pluralize(word, nb):
         else:
             comment = ""
 
+        if not all(revision.contains(issue) for issue in issues if issue.is_publishable):
```
La0, line 149: is_publishable is a method, not a property. Your test will always be true here as it tests the presend of the method...

bot/code_review_bot/report/base.py
```diff
@@ -146,6 +146,9 @@ def pluralize(word, nb):
...
+            comment = comment.replace(":", ": (defects might be in the parent stack)")
```
La0, line 150: Please don't use replace, it makes no sense.

```suggestion
            comment += ": (defects might be in the parent stack)"
```

bot/code_review_bot/report/base.py
```diff
@@ -146,6 +146,11 @@ def pluralize(word, nb):
         else:
             comment = ""
 
+        if not all(
+            revision.contains(issue) for issue in issues if issue.is_publishable()
+        ):
+            comment += "(defects might be in the parent stack)\n"
```
junngo, line 152: A new line will occur if I revise it by referring to the review :)
iss_new_line
If it is okay with you, I think you can use it like this instead of replace.

marco-c, line 152: We should only add the comment when there actually is a parent revision in the stack, otherwise it'll be confusing.

marco-c: Instead of adding this text, maybe `Code analysis found 2 defects in the diff 42:` could become `Code analysis found 2 defects (some in a parent revision) in the diff 42:`

junngo: How can I check that parent revision is in the stack?
Instead, I checked issue been published through `issue.is_publishable()` function.
Could you give me directions?

marco-c: > How can I check that parent revision is in the stack?

The revision is built in bot/code_review_bot/revisions.py. Maybe the stack information could be available in the result from this call https://github.com/mozilla/code-review/blob/50ea2f7b782fb7cd81d3cc54d82d2aa4611971b8/bot/code_review_bot/revisions.py#L175, you'll need to check.

>  Instead, I checked issue been published through issue.is_publishable() function.

We should do both.

junngo: Hello, @marco-c 
Sorry for late. I try to look at `revision` contents to check parent revision.
But, When I look at `revision`, I couldn't check parent revision.
Please, refer to the content.

- revision values
code_review_revision_contents

marco-c:
Closing given the discussion in #553.

PR #562, opened 2020-03-30
Display build error as coverity name

La0:
It will display `Build Error` as lint name, when a Coverity issue is a build error (defaulting to `Coverity` otherwise).

I also added the stack to the build error reporting, now that we use a lint result, it can be shown without breaking the display

### DIFF
--- a/bot/code_review_bot/__init__.py
+++ b/bot/code_review_bot/__init__.py
@@ -102,6 +102,14 @@ class Issue(abc.ABC):
         line = f"line {self.line}" if self.line is not None else "full file"
         return f"{self.analyzer.name} issue {self.check}@{self.level.value} {self.path} {line}"
 
+    @property
+    def display_name(self):
+        """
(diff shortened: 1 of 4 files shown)


La0:
It's the same structure as on a normal error. It was simply not used before because that put a lot of infos in the unit result display, at the top of the page.

A build error payload is visible on the first item of [that coverity.json](https://firefoxci.taskcluster-artifacts.net/MQ-1EhpUTpCvrDiwpLnDEw/0/public/code-review/coverity.json). It's the same thing as other issues.

marco-c:
> It's the same structure as on a normal error. It was simply not used before because that put a lot of infos in the unit result display, at the top of the page.
> A build error payload is visible on the first item of that coverity.json. It's the same thing as other issues.

I'm not 100% sure we want to do this, it seems that in the build error case we would write the same thing twice?

PR #569, opened 2020-04-06
Use rs-parsepatch for clang-format

La0:
This allows several good things:
1. no merges to do, we can process the chunks directly
2. support fix reporting directly at the Issue level (so we can support other analyzers)
3. We can easily detect the first modified line in the chunk to be more precise when reporting on Phabricator
4. we could build easily a ully appliable patch from publishable issue as we have the chunks ! (in a followup)


.taskcluster.yml
```diff
@@ -98,12 +98,11 @@ tasks:
             deadline: {$fromNow: '1 hour'}
             payload:
               maxRunTime: 3600
-              image: "python:${python_version}-alpine"
+              image: "python:${python_version}"
```
marco-c, line 101: Can we use slim here too?

La0: I think so

PR #571, opened 2020-04-07
Use rs-parsepatch instead of old python version.

La0:


La0:
`rs_parsepatch.get_lines` does not provide line numbers when a file is added

bot/code_review_bot/revisions.py
```diff
@@ -269,14 +269,36 @@ def analyze_patch(self):
...
-            filename: diff.get("touched", []) + diff.get("added", [])
-            for filename, diff in patch.items()
+            stat["filename"]: _get_lines(stat)
+            for stat in patch_stats
         }
 
+        from pprint import pprint
+
+        pprint(patch_stats)
+        pprint(self.lines)
```
marco-c, line 300: remove debugging stuff

bot/code_review_bot/revisions.py
```diff
@@ -269,14 +269,36 @@ def analyze_patch(self):
...
+        def _get_lines(file_stats):
+            if file_stats["new"]:
+                # file addition
+                return [line[1] for line in file_stats["lines"]]
+            else:
+                # file modification
+                return sorted(
+                    set(
+                        [
+                            line[0] or line[1]
```
marco-c, line 284: Was the behavior with parsepatch the same? Was it returning the line number in the `before` state?

La0: Filed an issue on parsepatch: https://github.com/mozilla/pyo3-parsepatch/issues/18

I'm not sure who's to blame here

La0: I was able to use directly `get_lines` for most cases, and fall back to `get_diffs` for additions only.

bot/code_review_bot/revisions.py
```diff
@@ -269,13 +269,26 @@ def analyze_patch(self):
         assert isinstance(self.patch, str), "Invalid patch type"
 
         # List all modified lines from current revision changes
-        patch = Patch.parse_patch(self.patch, skip_comments=False)
-        assert patch != {}, "Empty patch"
+        # associated with their diffs
+        # rs-parsepatch guarantees the order is the same in both lists
+        patch_stats = zip(
+            rs_parsepatch.get_lines(self.patch), rs_parsepatch.get_diffs(self.patch),
```
marco-c, line 275: This is... super painful. We have to parse the patch twice this way.

La0:
I ran [that patch on the test revision](https://phabricator-dev.allizom.org/D1758), and it did not alter the issues published !

bot/code_review_bot/revisions.py
La0: I fixed the local patch, and expanded it to make it more readable. The code is way simpler now

PR #582, opened 2020-04-15
events: Use jemalloc to reduce RAM usage on Heroku

### LINKED ISSUE #573: Use jemalloc in the Heroku worker image
The improvements for bugbug were huge: https://github.com/mozilla/bugbug/issues/1421.

With enough tuning, we might be able to use Standard-1X (https://github.com/mozilla/bugbug/issues/1472).

La0:
Fixes #573 & clone of https://github.com/mozilla/bugbug/issues/1421

I'm pushing the PR to testing to test the memory reduction

### DIFF
--- a/events/docker/Dockerfile
+++ b/events/docker/Dockerfile
@@ -10,6 +10,6 @@ RUN /src/tools/docker/bootstrap-mercurial.sh
 
 # Setup jemalloc to manage Ram more efficiently on Heroku
 ENV LD_PRELOAD="libjemalloc.so.2"
-ENV MALLOC_CONF="narenas:1"
+ENV MALLOC_CONF="narenas:1,tcache:false,background_thread:false,dirty_decay_ms:0,muzzy_decay_ms:0"
 
 CMD ["code-review-events"]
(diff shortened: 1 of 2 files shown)


events/docker/Dockerfile
```diff
@@ -8,4 +8,7 @@ RUN cd /src/events && python setup.py install
 
 RUN /src/tools/docker/bootstrap-mercurial.sh
 
+# Setup jemalloc to manage Ram more efficiently on Heroku
+ENV LD_PRELOAD="libjemalloc.so.2"
+
```
marco-c, line 13: Could you try to set narenas to 2 and see if that makes a difference?
`export MALLOC_CONF="narenas:2"`

marco-c: Or even 1, since we have a single thread

La0: Done in a49ce55 & pushing to testing...

La0:
Works on testing, with narenas=1, almost same result (~20% reduction in ram usage)

La0:
Added recommended options to the env variable, memory usage at 1180Mb instead of 1.5Gb during clone.

PR #601, opened 2020-05-04
bot: For C++ build error reporting stop using `Coverity` that proved to be suboptimal and use `clang-tidy` for this.

abpostelnicu:
`Coverity` proved to be suboptimal when we want to detect build errors so we must stop using it and switch to `clang-tooling`, in this case `clang-tidy`.

### DIFF
--- a/bot/code_review_bot/tasks/clang_tidy.py
+++ b/bot/code_review_bot/tasks/clang_tidy.py
@@ -40,6 +40,11 @@ ISSUE_NOTE_MARKDOWN = """
 ```
 """
 
+ERROR_MARKDOWN = """
+**Message**: ```{message}```
+**Location**: {location}
+"""
+
 CLANG_MACRO_DETECTION = re.compile(r"^expanded from macro")
 
 
@@ -57,6 +62,7 @@ class ClangTidyIssue(Issue):
         column,
         check,
         message,
+        level=Level.Warning,
         reliability=Reliability.Unknown,
         reason=None,
         publish=True,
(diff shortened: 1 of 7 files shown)


bot/code_review_bot/tasks/clang_tidy.py
```diff
@@ -78,6 +85,29 @@ def __init__(
         self.reliability = reliability
         self.publishable_check = publish
         self.reason = reason
+        self.diag_type = diag_type
```
marco-c, line 88: We could just use the already existing `level` attribute, why did you introduce this new attribute?

bot/code_review_bot/tasks/coverity.py
```diff
@@ -139,7 +133,7 @@ def as_markdown(self):
             publishable=self.is_publishable() and "yes" or "no",
             is_local=self.is_local() and "yes" or "no",
             reliability=self.reliability.value,
-            is_clang_error=self.is_clang_error() and "yes" or "no",
+            is_clang_error=self.is_build_error() and "yes" or "no",
```
marco-c, line 136: Nit: It seems we are not really using this `is_clang_error` anywhere, maybe you could remove it

abpostelnicu: Sure we can remove it.

bot/code_review_bot/tasks/clang_tidy.py
abpostelnicu: I wanted to better state that we are dealing there with a `diagnostic message`, but if you think there is no need to be extra explicit I'm fine removing it.

marco-c: Yeah I'd say better to simplify

PR #603, opened 2020-05-07
bot: Always publish build errors detected by `clang-tidy`.

abpostelnicu:


bot/code_review_bot/tasks/clang_tidy.py
```diff
@@ -85,6 +85,10 @@ def __init__(
         self.publishable_check = publish
         self.reason = reason
 
+        # For build errors always publish
+        if self.is_build_error():
+            self.publishable_check = True
```
marco-c, line 90: Isn't this already the case in the clang-tidy.json artifact?

abpostelnicu: Actually no, this is the [artifact](https://phabricator.services.mozilla.com/D74205#C2506775NL124). We don't publish a flag in the artifact because `clang-diagnostic-error` is not a checker from `clang-tidy` so we don't keep it in the `clang-tidy` [configuration](https://searchfox.org/mozilla-central/source/tools/clang-tidy/config.yaml) yaml. I've thought of adding it there, but it doesn't feel right since there are only checkers implemented in `clang-tidy`.

PR #647, opened 2020-07-09
bot: Add `--outgoing` for `clang-tidy` running message.

abpostelnicu:
Sometimes the issue detected by `clang-tidy` are in header files and thus the message that is displayed in `phabricator` includes those headers paths, this is something not useful sine they are not compilation units the message is actionable.

### DIFF
--- a/bot/code_review_bot/tasks/clang_tidy.py
+++ b/bot/code_review_bot/tasks/clang_tidy.py
@@ -184,8 +184,7 @@ class ClangTidyTask(AnalysisTask):
         return "clang-tidy"
 
     def build_help_message(self, files):
-        files = " ".join(files)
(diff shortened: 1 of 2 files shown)


bot/code_review_bot/tasks/clang_tidy.py
```diff
@@ -184,8 +184,7 @@ def display_name(self):
         return "clang-tidy"
 
     def build_help_message(self, files):
```
marco-c, line 186: I wonder if we are using the files parameter in other `build_help_message` functions. If not, we could remove it everywhere.

abpostelnicu: We are using it for `infer` and `clang-format`.

PR #684, opened 2020-08-05
bot: reviewbot post comments on phab revision everytime doc-upload task is ran

Emilfs:
Made changes to the reviewbot that makes it post a comment on a phab revision whenever doc-upload task is ran. The comment contains a link to an uploaded doc.


bot/tests/test_reporter_phabricator.py
```diff
@@ -25,6 +25,8 @@
 from code_review_bot.tasks.infer import InferTask
 from code_review_bot.tasks.lint import MozLintIssue
 from code_review_bot.tasks.lint import MozLintTask
+from code_review_bot.tasks.upload import DocUploadIssue
+from code_review_bot.tasks.upload import DocUploadTask
```
marco-c, line 29: You'll actually need to write some tests here, not just import the classes

bot/tests/test_workflow.py
```diff
@@ -16,6 +16,7 @@
 from code_review_bot.tasks.default import DefaultTask
 from code_review_bot.tasks.infer import InferTask
 from code_review_bot.tasks.lint import MozLintTask
+from code_review_bot.tasks.upload import DocUploadTask
```
marco-c, line 19: Similarly here, you'll actually need to write some tests, not just import the classes

bot/code_review_bot/workflow.py
```diff
@@ -322,6 +326,12 @@ def _in_group(dep_id):
                     continue
                 artifacts = task.load_artifacts(self.queue_service)
                 if artifacts is not None:
+                    if 'public/firefox-source-docs-url.txt' in artifacts:
+                        link_to_doc = artifacts['public/firefox-source-docs-url.txt'].decode('UTF-8')
+                        self.phabricator.comment(
```
marco-c, line 331: Maybe it should be part of the summary comment rather than a separate comment, wdyt @sylvestre?

bot/code_review_bot/workflow.py
```diff
@@ -25,12 +25,16 @@
...
 from code_review_bot.tasks.lint import MozLintTask
+from code_review_bot.tasks.upload import DocUploadTask
 
 logger = structlog.get_logger(__name__)
 
 TASKCLUSTER_NAMESPACE = "project.relman.{channel}.code-review.{name}"
 TASKCLUSTER_INDEX_TTL = 7  # in days
 
+COMMENT_LINK_TO_DOC = """
+We think you might have touched the doc files, generated doc can be accessed [here]({link_to_doc}).
```
marco-c, line 36: I'd write something like:
```suggestion
I think you might have touched the documentation, so I rendered it for you [here]({link_to_doc}).
```

marco-c: Or:
```suggestion
You have touched the documentation, you can find it rendered [here]({link_to_doc}).
```

bot/code_review_bot/tasks/upload.py
```diff
@@ -0,0 +1,91 @@
...
+- **Line**: {line}
+- **Publishable**: {publishable}
+
+```
+{message}
+```
+"""
+
+
+class DocUploadIssue(Issue):
```
marco-c, line 27: This class is exactly the same as DefaultIssue, you can just use it instead

bot/code_review_bot/tasks/upload.py
```diff
@@ -0,0 +1,91 @@
...
+    def display_name(self):
+        return "doc-upload"
+
+    def parse_issues(self, artifacts, revision):
+        artifact = artifacts.get("public/firefox-source-docs-url.txt")
+        if artifact is None:
+            logger.warn("Missing firefox-source-docs-url.txt")
+            return []
+
+        return [
```
marco-c, line 73: Can the source-test-doc-upload generate static analysis issues? I'd expect it can't and so we can just return an empty list here

Emilfs:
@marco-c ready fo review, please review :)

bot/code_review_bot/tasks/upload.py
```diff
@@ -0,0 +1,81 @@
...
+            logger.warn("Missing firefox-source-docs-url.txt")
+
+        return []
+
+    def get_link(self, artifacts):
+        artifact = artifacts.get("public/firefox-source-docs-url.txt")
+        if artifact is None:
+            logger.warn("Missing firefox-source-docs-url.txt")
+            return []
+
```
sylvestre, line 76: please avoid code duplication.
please create a new method doing that and call it from get_link and parse_issues

bot/code_review_bot/workflow.py
sylvestre: please take in account Marco's comment.
It could be:
You have touched the documentation, you can find it rendered [here]({link_to_doc}) for a week.

bot/code_review_bot/workflow.py
```diff
@@ -322,6 +334,11 @@ def _in_group(dep_id):
                     continue
                 artifacts = task.load_artifacts(self.queue_service)
                 if artifacts is not None:
+                    if "public/firefox-source-docs-url.txt" in artifacts:
```
sylvestre, line 337: I think you should declare "public/firefox-source-docs-url.txt" once as a member
in case the file name changes, we will need to update it only twice

bot/code_review_bot/tasks/upload.py
```diff
@@ -0,0 +1,69 @@
...
+- **Line**: {line}
+- **Publishable**: {publishable}
+
+```
+{message}
+```
+"""
+
+
+class DefaultIssue(Issue):
```
marco-c, line 24: This class doesn't seem necessary, you can remove it.

marco-c: That is, you are not using it anywhere :D

bot/code_review_bot/tasks/upload.py
```diff
@@ -0,0 +1,69 @@
...
+            level=self.level.value,
+            line=self.line,
+            message=self.message,
+            publishable=self.is_publishable() and "yes" or "no",
+        )
+
+
+class DocUploadTask(AnalysisTask):
+    """
+    Support issues from source-test clang-format tasks by reading the
```
marco-c, line 54: The comment needs to be changed, it refers to clang-format

bot/code_review_bot/tasks/upload.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: The name of the file, `upload.py`, is not great. I can't think of a super nice name, but `docs.py` or `docupload.py` would definitely be better.

@sylvestre any idea?

bot/code_review_bot/workflow.py
```diff
@@ -188,7 +196,10 @@ def publish(self, revision, issues, task_failures):
         # Publish reports about these issues
         with stats.timer("runtime.reports"):
             for reporter in self.reporters.values():
-                reporter.publish(issues, revision, task_failures)
+                if isinstance(reporter, PhabricatorReporter):
+                    reporter.publish(issues, revision, task_failures, link_to_doc)
+                else:
+                    reporter.publish(issues, revision, task_failures)
```
marco-c, line 202: Instead of using a different `publish` function according to the class, implement the same `publish` signature in all reporter classes.

bot/code_review_bot/workflow.py
```diff
@@ -322,6 +335,9 @@ def _in_group(dep_id):
                     continue
                 artifacts = task.load_artifacts(self.queue_service)
                 if artifacts is not None:
+                    if docupload_artifact_path in artifacts:
+                        link_to_doc = artifacts[docupload_artifact_path].decode("UTF-8")
```
marco-c, line 339: This code is very specific to docs, so it doesn't belong to the generic `workflow.py`, but it should be in the DocUploadTask class.

You can add a function such as `build_link` in bot/code_review_bot/tasks/base.py which returns `""` by default, and override it in DocUploadTask to return the link to the docs.
Then, in workflow.py you'd simply call `task.build_link`.

bot/code_review_bot/tasks/docupload.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+import structlog
+
+from code_review_bot.tasks.base import AnalysisTask
+
+logger = structlog.get_logger(__name__)
+
+ISSUE_MARKDOWN = """
```
marco-c, line 8: This isn't used anywhere, you can remove it.

bot/code_review_bot/tasks/docupload.py
```diff
@@ -0,0 +1,48 @@
...
+    artifacts = ["public/firefox-source-docs-url.txt"]
+
+    @property
+    def display_name(self):
+        return "doc-upload"
+
+    def parse_issues(self, artifacts, revision):
+        artifact = artifacts.get("public/firefox-source-docs-url.txt")
+        if artifact is None:
+            logger.warn("Missing firefox-source-docs-url.txt")
```
marco-c, line 37: No need to get the artifact, since you're not using it and you are returning an empty list anyway at the end.

bot/code_review_bot/tasks/docupload.py
```diff
@@ -0,0 +1,48 @@
...
+
+        return []
+
+    def build_link(self, artifacts):
+        artifact = artifacts.get("public/firefox-source-docs-url.txt")
+        if artifact is None:
+            logger.warn("Missing or empty firefox-source-docs-url.txt")
+            return []
```
marco-c, line 45: `build_link` is supposed to return a string, so this should be `return ""`.

bot/code_review_bot/tasks/docupload.py
```diff
@@ -0,0 +1,48 @@
...
+
+        link_to_doc = artifact.decode("utf-8")
+        return link_to_doc
```
marco-c, line 48: Nit: replace these two lines with a single `return artifact.decode("utf-8")`.

bot/code_review_bot/workflow.py
```diff
@@ -310,6 +317,7 @@ def _in_group(dep_id):
         # Find issues and patches in dependencies
         issues = []
         task_failures = []
+        link_to_doc = ""
```
marco-c, line 320: Let's make this more generic, assuming any analyzer could return a link.
So, make `link_to_doc` a list.

bot/code_review_bot/workflow.py
```diff
@@ -335,6 +343,9 @@ def _in_group(dep_id):
                     for patch in task_patches:
                         revision.add_improvement_patch(task, patch)
 
+                    if task.build_link(artifacts):
+                        link_to_doc = task.build_link(artifacts)
```
marco-c, line 347: And here do something like:
```
link = task.build_link(artifacts)
if link:
    links.append(link)
```

bot/code_review_bot/workflow.py
marco-c: Don't forget this!

bot/code_review_bot/report/base.py
```diff
@@ -183,6 +187,9 @@ def pluralize(word, nb):
         if CoverageIssue in issue_classes:
             comment += COMMENT_COVERAGE
 
+        if links:
+            comment += COMMENT_LINK_TO_DOC.format(link_to_doc=links[0])
```
marco-c, line 191: Add a comment here like this:
```
# For now, we assume there is only one task that creates a link, the doc-upload task.
```

And an assertion such as:
```
assert len(links) == 1
```

bot/code_review_bot/report/phabricator.py
```diff
@@ -105,7 +105,7 @@ def publish_harbormaster(
             nb_unit=len(unit_issues),
         )
 
-    def publish_summary(self, revision, issues, patches, task_failures):
+    def publish_summary(self, revision, issues, patches, task_failures, links=None):
```
marco-c, line 108: Nit: Probably the default `=None` is no longer necessary, could you try?

bot/code_review_bot/workflow.py
```diff
@@ -301,7 +302,7 @@ def _in_group(dep_id):
             task = tasks[dependencies[0]]
             if task["task"]["metadata"]["name"] == "Gecko Decision Task":
                 logger.warn("Only dependency is a Decision Task, skipping analysis")
-                return [], []
+                return [], [], ""
```
marco-c, line 305: The third should now be `[]` instead of `"`, since links is a list

bot/tests/test_remote.py
```diff
@@ -64,8 +64,8 @@ def test_baseline(mock_config, mock_revision, mock_workflow, mock_backend, mock_
     - an analyzer in failed status
     - with some issues in its log
     """
-    from code_review_bot.tasks.lint import MozLintIssue
     from code_review_bot.tasks.coverage import CoverageIssue
+    from code_review_bot.tasks.lint import MozLintIssue
```
marco-c, line 68: This change (done by pre-commit's isort) should already be on master, if you rebase (`git pull upstream master --rebase`) it will be picked up.

marco-c (commented):
It's looking good, almost there!

Emilfs:
never tried rebase before, my number of commits doubled now, is it normal?

sylvestre:
git rebase origin/master 
and a force push (git push -f) should fix it

bot/code_review_bot/workflow.py
sylvestre: you forgot the "for a week" in the commit message :)
it is important to highlight that, these urls will expire

marco-c:
Looks good, the only thing to change is what Sylvestre suggested ("You have touched the documentation, you can find it rendered [here]({link_to_doc}) for a week" instead of "You have touched the documentation, you can find it rendered [here]({link_to_doc})").

marco-c (approved):

marco-c:
Thanks a lot for your work!

PR #690, opened 2020-08-15
Add a comment when errors are published outside of patch

### LINKED ISSUE #553: Add a comment when errors are published outside of patch

bhushan-borole:
Fixes #553

### DIFF
--- a/bot/code_review_bot/report/base.py
+++ b/bot/code_review_bot/report/base.py
@@ -158,6 +158,11 @@ class Reporter(object):
         else:
             comment = ""
 
+        if not all(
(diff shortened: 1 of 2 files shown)


marco-c (changes requested):
This is the same as #561, the same comments apply.

marco-c:
Closing this PR since it's stale.

PR #718, opened 2020-09-25
Add support for source-test-clang-tidy-external

mozfreddyb:
This adds support for new static analysis, called "source-test-clang-tidy-external", which we are enabling in [Bug 1666808](https://bugzilla.mozilla.org/show_bug.cgi?id=1666808) (see also changeset in [phab](https://phabricator.services.mozilla.com/D91149)).

It works locally and pytest & flake8 pass, but I'm sure there's going to be some iteration.

The reported message looks like this, happy to modify to make it work for you folks.
image

### DIFF
--- a/bot/code_review_bot/report/base.py
+++ b/bot/code_review_bot/report/base.py
@@ -5,6 +5,7 @@
 
 import itertools
 
+from code_review_bot.tasks.clang_tidy_external import ExternalTidyIssue
 from code_review_bot.tasks.coverage import CoverageIssue
 from code_review_tools import treeherder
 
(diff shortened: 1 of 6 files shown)


mozfreddyb:
I'll acknowledge that the reporting looks a bit odd, but I did not want to break existing templates too much.
Unfortunately, the fact that it currently expects instructions for how to reproduce locally does not combine with _private_ static analysis. I'm happy to keep iterating on that.

mozfreddyb:
@abpostelnicu When do you think will you be able to review this?

abpostelnicu:
Sorry, I will do it later today,

abpostelnicu (approved):

PR #744, opened 2020-10-16
Improve test coverage for clang-tidy-external

mozfreddyb:
This patch adds testing for the case in which clang-tidy-external does not report clang-diagnostic errors,
which ought to be caught by the other clang-tidy jobs or during built.

We also improve reporting logic to ensure that only `mozilla-civet`-prefixed
diagnotics are reported.

### DIFF
--- a/bot/code_review_bot/tasks/clang_tidy_external.py
+++ b/bot/code_review_bot/tasks/clang_tidy_external.py
@@ -42,10 +42,12 @@ ERROR_MARKDOWN = """
 
 CLANG_MACRO_DETECTION = re.compile(r"^expanded from macro")
 
+BUILD_HELP_MSG = """For private static analysis, please see [our private docs in Mana](https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=130909687), if you cannot access this resource, ask your reviewer to help you resolve the issue."""
+
 
 class ExternalTidyIssue(ClangTidyIssue):
     """
-    An issue reported by source-test-clang-tidy-external
+    An issue reported by source-test-clang-external
     """
 
     def as_error(self):
@@ -130,7 +132,7 @@ class ExternalTidyIssue(ClangTidyIssue):
 
 class ExternalTidyTask(ClangTidyTask):
     """
-    Support issues from source-test clang-tidy-external tasks
+    Support issues from source-test clang-external tasks
     """
 
     # Note this is currently in fact using the same file name as the
@@ -143,7 +145,7 @@ class ExternalTidyTask(ClangTidyTask):
         return "private static analysis"
 
     def build_help_message(self, files):
-        return "Unfortunately, private static analysis can not be reproduced locally."
+        return BUILD_HELP_MSG
 
     def parse_issues(self, artifacts, revision):
         issues = [
@@ -160,7 +162,8 @@ class ExternalTidyTask(ClangTidyTask):
                 if "reliability" in warning
                 else Reliability.Unknown,
                 reason=warning.get("reason"),
-                publish=warning.get("publish"),
+                publish=warning.get("publish")
+                and warning["flag"].startswith("mozilla-civet-"),
             )
             for artifact in artifacts.values()
             for path, items in artifact["files"].items()
--- a/bot/code_review_bot/workflow.py
+++ b/bot/code_review_bot/workflow.py
@@ -399,7 +399,7 @@ class Workflow(object):
             return InferTask(task_id, task_status)
         elif name == "source-test-doc-upload":
             return DocUploadTask(task_id, task_status)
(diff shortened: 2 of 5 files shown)


bot/code_review_bot/tasks/clang_tidy_external.py
```diff
@@ -160,7 +170,8 @@ def parse_issues(self, artifacts, revision):
                 if "reliability" in warning
                 else Reliability.Unknown,
                 reason=warning.get("reason"),
-                publish=warning.get("publish"),
+                publish=warning.get("publish")
+                and not warning["flag"].startswith("clang-diagnostic"),
```
marco-c, line 174: What about doing `warning["flag"].startswith("mozilla-civet-")` instead of `not warning["flag"].startswith("clang-diagnostic")`?

bot/code_review_bot/tasks/clang_tidy_external.py
```diff
@@ -127,6 +130,13 @@ def as_markdown(self):
             ),
         )
 
+    def is_publishable(self):
+        """
+        Is this issue publishable on reporters ?
+        """
+
+        return super().is_publishable and not self.check.startswith("clang-diagnostic")
```
marco-c, line 138: What about doing `self.check.startswith("mozilla-civet-")` instead of `not self.check.startswith("clang-diagnostic")`?

bot/code_review_bot/tasks/clang_tidy_external.py
```diff
@@ -127,6 +130,13 @@ def as_markdown(self):
             ),
         )
 
+    def is_publishable(self):
```
marco-c, line 133: You don't need to override `is_publishable`, you can override `validates` instead (and this way you don't have to manually check `super().is_publishable`).

marco-c: And actually, with the modification below (in the call to the ExternalTidyIssue constructor), you shouldn't need to override this.
`validates` in ClangTidyIssue checks if publishable_check is True, and publishable_check is the value you pass with `publish` argument.

mozfreddyb: gotcha. removing this bit then. thanks!

bot/code_review_bot/tasks/clang_tidy_external.py
```diff
@@ -42,10 +42,12 @@
 
 CLANG_MACRO_DETECTION = re.compile(r"^expanded from macro")
 
+BUILD_HELP_MSG = """For private static analysis, please see [our private docs in Mana](https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=130909687), if you cannot access this resource, ask your reviewer to help you resolve the issue."""
+
 
 class ExternalTidyIssue(ClangTidyIssue):
     """
-    An issue reported by source-test-clang-tidy-external
+    An issue reported by source-test-clang-external
```
marco-c, line 50: Isn't it `source-test-clang-tidy-external`?

bot/code_review_bot/tasks/clang_tidy_external.py
marco-c: @mozfreddyb this is still there :)

bot/code_review_bot/tasks/clang_tidy_external.py
mozfreddyb: Unfortunately not. It is "clang-tidy-external" and "source-test-clang-external". :confused:

abpostelnicu, line 50: Do we really want this change? For me this is sounds like an external compiler set based on clang and not clang-tidy based static-analysis.

mozfreddyb:
That’s already the name of the job. I’m just aligning with task name (and
job kind). Marco and I discussed this and we agreed to fix naming in a
follow up. - ok?

ANdi <notifications@github.com> schrieb am Do. 12. Nov. 2020 um 17:35:

abpostelnicu:
> That’s already the name of the job. I’m just aligning with task name (and job kind). Marco and I discussed this and we agreed to fix naming in a follow up. - ok? ANdi <notifications@github.com> schrieb am Do. 12. Nov. 2020 um 17:35:
> […](#)

I'm sorry that I have not been part of that discussion, for me the current naming scheme "clang-external" is a little bit ambiguous. But in the end it's your decision here.

mozfreddyb:
We’ll fix it in a follow up. I agree the name is confusing but that’s
already established in taskcluster. I’m filing a follow up bug tomorrow.

ANdi <notifications@github.com> schrieb am Do. 12. Nov. 2020 um 17:50:

abpostelnicu:
Thank you so much!

marco-c (approved):

PR #792, opened 2020-12-07
backend: For the postgresql `DiffViewSet` do not use historic diffs older than 3 months.

abpostelnicu:
Because of performance issues that is caused by a great number of entries in the `issues.issue` table we should not look at historic data that is not that relevant.

### DIFF
--- a/backend/code_review_backend/issues/api.py
+++ b/backend/code_review_backend/issues/api.py
@@ -3,7 +3,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from datetime import date
 from datetime import datetime
+from datetime import timedelta
 
 from django.db.models import Count
 from django.db.models import Q
@@ -93,7 +95,9 @@ class DiffViewSet(viewsets.ReadOnlyModelViewSet):
 
     def get_queryset(self):
         diffs = (
-            Diff.objects.all()
+            Diff.objects
+            # Because of the perf. hit filter issues that are not older than today - 3 months.
+            .filter(created__gte=date.today() - timedelta(days=90))
             .prefetch_related(
                 "issues", "revision", "revision__repository", "repository"
             )


sylvestre:
The cost of the old query:
```
    - Plan: 
        Node Type: "Aggregate"
        Strategy: "Plain"
        Partial Mode: "Simple"
        Parallel Aware: false
        Startup Cost: 437486.67
        Total Cost: 437486.67
        Plan Rows: 1
        Plan Width: 8
        Plans: 
          - Node Type: "Aggregate"
            Strategy: "Hashed"
            Partial Mode: "Simple"
            Parent Relationship: "Outer"
            Parallel Aware: false
            Startup Cost: 436590.07
            Total Cost: 437003.88
            Plan Rows: 137939
            Plan Width: 36
            Group Key: 
              - "issues_diff.id"
            Plans: 
              - Node Type: "Hash Join"
                Parent Relationship: "Outer"
                Parallel Aware: false
                Join Type: "Right"
                Startup Cost: 3549.60
                Total Cost: 432551.40
                Plan Rows: 8077335
                Plan Width: 4
                Inner Unique: true
                Hash Cond: "(issues_issue.diff_id = issues_diff.id)"
                Plans: 
                  - Node Type: "Seq Scan"
                    Parent Relationship: "Outer"
                    Parallel Aware: false
                    Relation Name: "issues_issue"
                    Alias: "issues_issue"
                    Startup Cost: 0.00
                    Total Cost: 424761.01
                    Plan Rows: 8077335
                    Plan Width: 4
                  - Node Type: "Hash"
                    Parent Relationship: "Inner"
                    Parallel Aware: false
                    Startup Cost: 3066.82
                    Total Cost: 3066.82
                    Plan Rows: 137939
                    Plan Width: 4
                    Plans: 
                      - Node Type: "Seq Scan"
                        Parent Relationship: "Outer"
                        Parallel Aware: false
                        Relation Name: "issues_diff"
                        Alias: "issues_diff"
                        Startup Cost: 0.00
                        Total Cost: 3066.82
                        Plan Rows: 137939
                        Plan Width: 4
```

abpostelnicu:
@marco-c if you have any suggestion here, you are more than welcome to comment/review.

backend/code_review_backend/issues/api.py
```diff
@@ -93,7 +95,9 @@ class DiffViewSet(viewsets.ReadOnlyModelViewSet):
 
     def get_queryset(self):
         diffs = (
-            Diff.objects.all()
+            Diff.objects
+            # Because of the perf. hit filter issues that are not older than today - 3 months.
+            .filter(created__gte=date.today() + relativedelta(months=-3))
```
marco-c, line 100: Nit: you could use the standard Python lib here to avoid adding `dateutil` to the dependencies. It won't be precisely three months, but it should be OK for our needs: `datetime.now() - timedelta(days=90)`

marco-c (dismissed):

backend/code_review_backend/issues/api.py
```diff
@@ -93,7 +95,9 @@ class DiffViewSet(viewsets.ReadOnlyModelViewSet):
 
     def get_queryset(self):
         diffs = (
-            Diff.objects.all()
+            Diff.objects
+            # Because of the perf. hit filter issues that are not older than today - 3 months.
+            .filter(created__gte=date.today() + timedelta(days=90))
```
marco-c, line 100: It should be `- timedelta`, otherwise you'll only look for "future" issues :P

abpostelnicu: Yes, I've seen it after I've pushed it 👍

marco-c (approved):

PR #835, opened 2021-02-18
backend: For the `IssueCheckStats` do not use historic diffs older than 3 months.

abpostelnicu:
Because of the multiple joins that this feature uses and the inherent performance hit we should stick with results that are in the timespan of `today - 90 days`.


abpostelnicu:
@marco-c there are some issues when generating the docker image files.

marco-c (approved):

marco-c:
> @marco-c there are some issues when generating the docker image files.

We had the same issue in bugzilla-dashboard-backend, it can be fixed by adding an explicit dependency for google-auth (see https://github.com/mozilla/bugzilla-dashboard-backend/pull/22/files).

PR #839, opened 2021-02-19
events: Add an explicit dep for google-auth

abpostelnicu:


marco-c:
No longer needed after the libmozdata upgrade.

PR #858, opened 2021-03-09
clang-format: Suggest to directly modify the file instead

mozfreddyb:


mozfreddyb:
Should probably follow up with an email to dev-platform or such, if people relied on the other thing?

marco-c (commented):
There are some failing tests

> Should probably follow up with an email to dev-platform or such, if people relied on the other thing?

Yeah, why not, it doesn't hurt :)

abpostelnicu (approved):

mozfreddyb:
@marco-c good to go?

PR #974, opened 2021-09-08
events: in case of SIGTERM reset build to libmozevent in case it wasn't sent

abpostelnicu:


abpostelnicu:
Of course this can be also moved as part of Events.

events/code_review_events/workflow.py
```diff
@@ -397,6 +412,18 @@ def __init__(self, cache_root):
...
+        if isinstance(self.workflow, CodeReview) is False:
+            return
+
+        logger.info("Events: received SIGTERM")
+        if self.workflow.cached_build() is not None:
+            logger.info(
+                "Events: having unprocessed build",
+                build=str(self.workflow.cached_build()),
+            )
+            self.workflow.process_build(self.workflow.cached_build())
```
marco-c, line 425: In SIGTERM handler, we have limited time to cleanup. Instead of trying to process the build, we should just put it back in the queue, so the next worker that boots up will process it.

marco-c:
One problem we have if we want to fully fix this is that in `process_build` there is:
```
            # Then send the build toward next stage
            logger.info("Send build to Mercurial", build=str(build))
            await self.bus.send(QUEUE_MERCURIAL, build)
            # Report public bug as 'working' (in progress)
            await self.bus.send(QUEUE_PHABRICATOR_RESULTS, ("work", build, {}))
            # Send to bugbug workflow
await self.bus.send(QUEUE_BUGBUG, build)
```

What if we get terminated after we sent only one or two? We should send them all or none. We could add support for transactions in the MessageBus class.

events/code_review_events/workflow.py
abpostelnicu: What does limited time mean? in POSIX there is no standard telling us that there is such a time between SIGTERM and SIGKILL or SIGSTOP. 
SIGTERM only signals our app to stop but since we don't intercept that signal we don't stop and that maybe the send message is sent.

abpostelnicu: I think we should intercept SIGTERM, do the cleaning and terminate the ap ourselfs.

marco-c: Heroku sends a SIGTERM and gives you 30 seconds to cleanup.

abpostelnicu: This is great then! But if we terminate the app in the callback do you think that there is the possibility of not consuming another message from the bus?

marco-c: It's OK if we don't consume the messages, as they will stay in the queue and the next worker that is going to be started is goint to consume them properly.

PR #975, opened 2021-09-10
Add lando warnings support

abpostelnicu:
With https://github.com/mozilla-conduit/lando-api/pull/172 we have support in lando to send warnings.
We are going to send two warnings for each diff for each revision:

- Once the revision has been updated/created an initial message will be send indicating the user that there are linters running:"Review time linters are still running. Are you sure you want to continue?"
- Just as the review bot finishes posting issues in phab we send the 2nd message with the issues or if there aren't any we clear previous send message.

### DIFF
--- a/bot/code_review_bot/report/lando.py
+++ b/bot/code_review_bot/report/lando.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import structlog
+
+from code_review_bot import Level
+from code_review_bot.report.base import Reporter
+
+logger = structlog.get_logger(__name__)
+
+LANDO_MESSAGE = "The code review bot found {errors} {errors_noun} which should be fixed to avoid backout and {warnings} {warnings_noun}."
+
+
+class LandoReporter(Reporter):
+    """
+    Update lando with a warning message
+    """
+
+    def __init__(self, configuration):
+        self.lando_api = None
+
+    def setup_api(self, lando_api):
+        logger.info("Publishing warnings to lando is enabled by the bot!")
+        self.lando_api = lando_api
+
+    def publish(self, issues, revision, task_failures, links):
+        """
+        Send an email to administrators
+        """
+        if self.lando_api is None:
+            logger.info("Lando integration is not set!")
+            return
+
+        nb_publishable = len([i for i in issues if i.is_publishable()])
+        nb_publishable_errors = sum(
+            1 for i in issues if i.is_publishable() and i.level == Level.Error
+        )
+
+        nb_publishable_warnings = nb_publishable - nb_publishable_errors
+
+        logger.info(
+            "Publishing warnings to lando for {0} errors and {1} warnings".format(
+                nb_publishable_errors, nb_publishable_warnings
+            ),
+            revision=revision.id,
+            diff=revision.diff["id"],
+        )
+
+        try:
+            # code-review.events sends an initial warning message to lando to specify that the analysis is in progress,
+            # we should remove it
+            self.lando_api.del_all_warnings(revision.id, revision.diff["id"])
+
+            if nb_publishable > 0:
+                self.lando_api.add_warning(
+                    LANDO_MESSAGE.format(
+                        errors=nb_publishable_errors,
+                        errors_noun="error" if nb_publishable_errors == 1 else "errors",
+                        warnings=nb_publishable_warnings,
+                        warnings_noun="warning"
+                        if nb_publishable_warnings == 1
+                        else "warnings",
+                    ),
+                    revision.id,
+                    revision.diff["id"],
+                )
+        except Exception as ex:
+            logger.error(str(ex))
--- a/events/code_review_events/workflow.py
+++ b/events/code_review_events/workflow.py
@@ -2,6 +2,7 @@
 import asyncio
 
 import structlog
+from libmozdata.lando import LandoWarnings
 from libmozdata.phabricator import BuildState
 from libmozdata.phabricator import UnitResult
 from libmozdata.phabricator import UnitResultState
@@ -39,6 +40,14 @@ PULSE_TASK_GROUP_RESOLVED = "exchange/taskcluster-queue/v1/task-group-resolved"
 PULSE_TASK_COMPLETED = "exchange/taskcluster-queue/v1/task-completed"
 PULSE_TASK_FAILED = "exchange/taskcluster-queue/v1/task-failed"
 
+LANDO_WARNING_MESSAGE = "Static analysis and linting are still in progress."
+LANDO_FAILURE_MESSAGE = (
+    "Static analysis and linting did not run due to a generic failure."
+)
+LANDO_FAILURE_HG_MESSAGE = (
+    "Static analysis and linting did not run due to failure in applying the patch."
+)
+
 
 class CodeReview(PhabricatorActions):
     """
@@ -46,7 +55,15 @@ class CodeReview(PhabricatorActions):
     and pushing on Try repositories
     """
 
-    def __init__(self, publish=False, user_blacklist=[], *args, **kwargs):
+    def __init__(
+        self,
+        lando_url,
+        lando_publish_generic_failure,
+        publish=False,
+        user_blacklist=[],
+        *args,
+        **kwargs,
+    ):
         super().__init__(*args, **kwargs)
         self.publish = publish
         logger.info(
@@ -54,6 +71,17 @@ class CodeReview(PhabricatorActions):
                 self.publish and "enabled" or "disabled"
             )
         )
+
+        self.publish_lando = True if lando_url is not None else False
+        self.lando_publish_failure = lando_publish_generic_failure
+        if self.publish_lando:
+            logger.info("Publishing warnings to lando is enabled!")
+            self.lando_warnings = LandoWarnings(
+                api_url=lando_url,
+                api_key=kwargs["api_key"],
+            )
+            if self.lando_publish_failure:
+                logger.info("Publishing all failures to lando is enabled!")
 
         # Load the blacklisted users
         if user_blacklist:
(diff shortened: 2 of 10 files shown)


abpostelnicu:
This has been thoroughly tested on testing and it works like a charm. Unfortunately repo `mozilla-central` is not compatible with the new feature from lando and the warning messages cannot be displayed in the UI bu they appear in the backend.

abpostelnicu:
@zzzeid do you think we can add `mozilla-central` to be compatible on lando testing with the new feature set?

bot/code_review_bot/workflow.py
```diff
@@ -79,6 +80,9 @@ def __init__(
         # Setup Backend API client
         self.backend_api = BackendAPI()
 
+        # Setting up lando
+        self.lando = lando
```
marco-c, line 84: Nit: `lando_api`

bot/code_review_bot/workflow.py
```diff
@@ -191,6 +197,41 @@ def publish(self, revision, issues, task_failures, links):
         )
         stats.add_metric("analysis.issues.publishable", nb_publishable)
 
+        if self.lando is not None:
+            logger.info(
+                "Publishing warnings to lando for {0} errors and {1} warnings".format(
+                    nb_publishable_errors, nb_publishable - nb_publishable_errors
```
marco-c, line 203: Nit: replace `nb_publishable - nb_publishable_errors` with `nb_publishable_warnings`

bot/code_review_bot/workflow.py
```diff
@@ -191,6 +197,41 @@ def publish(self, revision, issues, task_failures, links):
...
+                ),
+                revision=revision.id,
+                diff=revision.diff["id"],
+            )
+            self.lando.del_all_warnings(revision.id, revision.diff["id"])
```
marco-c, line 208: Can you add a comment explaining that we delete all warnings here to remove the warning related to the build being in progress?

bot/code_review_bot/workflow.py
```diff
@@ -191,6 +197,41 @@ def publish(self, revision, issues, task_failures, links):
...
+                    + (
+                        " {} warnings.".format(nb_publishable_warnings)
+                        if nb_publishable_warnings
+                        else ""
+                    )
+                    + " Are you sure you want to continue?"
+                )
+                self.lando.add_warning(
+                    lando_warning_message, revision.id, revision.diff["id"]
+                )
```
marco-c, line 233: Instead of having this code here, it'd be cleaner to build a new `Reporter` class (like we have the `PhabricatorReporter`), e.g. `LandoReporter`.

abpostelnicu: I thought about that, but we would have a special case when no matter if we have or don't have issues pending we will need to do reporter.publish()

bot/code_review_bot/workflow.py
```diff
@@ -191,6 +197,41 @@ def publish(self, revision, issues, task_failures, links):
...
+                    nb_publishable_errors, nb_publishable - nb_publishable_errors
+                ),
+                revision=revision.id,
+                diff=revision.diff["id"],
+            )
+            self.lando.del_all_warnings(revision.id, revision.diff["id"])
+
+            # Determine if we need to post something or not
+            if nb_publishable > 0:
+                lando_warning_message = (
```
marco-c, line 212: Nit: Let's define the template at the top of the file, so it's more readable (like https://github.com/mozilla/code-review/blob/ef9459f2211cd9eca869e5ea60faecb508288bc1/bot/code_review_bot/report/mail.py#L17).

We should also make it clear as part of the message that errors will likely lead to a back-out.
We could write something like:
`The code review bot found XXX errors which should be fixed to avoid backout and XXX warnings`

tools/code_review_tools/lando.py
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Heroku Documentation: https://devcenter.heroku.com/
```
marco-c, line 6: Nit: this comment line is unrelated to this file

tools/code_review_tools/lando.py
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: Could you move this file to https://github.com/mozilla/libmozdata instead of adding it here?
We might want to use it in other projects too.

events/code_review_events/workflow.py
```diff
@@ -32,21 +32,26 @@
...
+from code_review_tools.lando import LandoWarnings
 
 logger = structlog.get_logger(__name__)
 
 PULSE_TASK_GROUP_RESOLVED = "exchange/taskcluster-queue/v1/task-group-resolved"
 PULSE_TASK_COMPLETED = "exchange/taskcluster-queue/v1/task-completed"
 PULSE_TASK_FAILED = "exchange/taskcluster-queue/v1/task-failed"
 
+LANDO_WARNING_MESSAGE = (
+    "Review time linters are still running. Are you sure you want to continue?"
```
marco-c, line 44: `Static analysis and linting is still in progress` (no need for the second part "Are you sure you want to continue?" as this part is already in the Lando UI itself)

bot/code_review_bot/workflow.py
marco-c: Yes, but that's not a problem, the PhabricatorReporter also does that check

marco-c:
Could you write some tests for this feature in `bot` and `events`?

tools/code_review_tools/lando.py
abpostelnicu: OK, I will move it to libmozdada

bot/code_review_bot/report/lando.py
```diff
@@ -0,0 +1,65 @@
...
+            if nb_publishable > 0:
+                self.lando_api.add_warning(
+                    LANDO_MESSAGE.format(
+                        errors=nb_publishable_errors, warnings=nb_publishable_warnings
+                    ),
+                    revision.id,
+                    revision.diff["id"],
+                )
+        except Exception as ex:
+            logger.info(str(ex))
```
marco-c, line 65: In this case we should log an error rather than an info message

bot/code_review_bot/report/lando.py
```diff
@@ -0,0 +1,65 @@
...
+        logger.info(
+            "Publishing warnings to lando for {0} errors and {1} warnings".format(
+                nb_publishable_errors, nb_publishable - nb_publishable_errors
+            ),
+            revision=revision.id,
+            diff=revision.diff["id"],
+        )
+
+        try:
+            # code-review.events sends an initial warning message to lando,
```
marco-c, line 52:
```suggestion
            # code-review.events sends an initial warning message to lando to specify that the analysis is in progress,
```

bot/code_review_bot/workflow.py
```diff
@@ -183,6 +183,7 @@ def publish(self, revision, issues, task_failures, links):
         nb_publishable_errors = sum(
             1 for i in issues if i.is_publishable() and i.level == Level.Error
         )
+
```
marco-c, line 186: Nit: please remove this extraneous newline

events/code_review_events/workflow.py
```diff
@@ -134,6 +145,20 @@ async def process_build(self, build):
...
+                    "Begin publishing init warning message to lando.",
+                    revision=build.revision["id"],
+                    diff=build.diff_id,
+                )
+                try:
+                    self.lando_warnings.add_warning(
+                        LANDO_WARNING_MESSAGE, build.revision["id"], build.diff_id
+                    )
+                except Exception as ex:
+                    logger.info(str(ex))
```
marco-c, line 160: Here too, we should log an error

marco-c:
Looks good overall, but we need to add some tests to both `bot` and `events`

bot/code_review_bot/report/lando.py
```diff
@@ -0,0 +1,65 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import structlog
+
+from code_review_bot import Level
+from code_review_bot.report.base import Reporter
+
+logger = structlog.get_logger(__name__)
+
+LANDO_MESSAGE = "The code review bot found {errors} errors which should be fixed to avoid backout and {warnings} warnings."
```
marco-c, line 13: Nit: we could use the singular form when there is a single error or a single warning

bot/tests/test_lando.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: Can you rename this file `test_reporter_lando.py`? This way we match the names of the other tests for the other reporters.

events/code_review_events/workflow.py
```diff
@@ -39,14 +40,30 @@
 PULSE_TASK_COMPLETED = "exchange/taskcluster-queue/v1/task-completed"
 PULSE_TASK_FAILED = "exchange/taskcluster-queue/v1/task-failed"
 
+LANDO_WARNING_MESSAGE = "Static analysis and linting is still in progress."
```
marco-c, line 43: Should it be
```suggestion
LANDO_WARNING_MESSAGE = "Static analysis and linting are still in progress."
```
?

I'm not sure which one is grammatically correct

events/code_review_events/workflow.py
```diff
@@ -39,14 +40,30 @@
...
+LANDO_FAILURE_MESSAGE = (
+    "Static analysis and linting did not run due to general failure."
```
marco-c, line 45:
```suggestion
    "Static analysis and linting did not run due to a generic failure."
```

events/tests/test_workflow.py
```diff
@@ -48,7 +80,9 @@ async def test_blacklist(PhabricatorMock, mock_taskcluster):
 
 
 @pytest.mark.asyncio
-async def test_publish_results_success_mode(PhabricatorMock, mock_taskcluster):
+async def test_publish_results_success_and_error_mode(
```
marco-c, line 83: Let's split this into multiple smaller tests instead of modifying the existing test, this way we know right away, in case of failure, what exactly is failing

abpostelnicu: Cool I'm going to do 3 tests, lando on success, fail and general_fail.

marco-c (approved):

PR #986, opened 2021-09-23
bot: when a generic bot failure occurs also report it in phabricator.

### LINKED ISSUE #984: When failing during workflow.run, post a generic failure like we do during events
When an exception is thrown during the bot workflow, we report a build failure without any additional information: https://github.com/mozilla/code-review/blob/71390ca1bf455fde5db4b34dc271307edf5bb955/bot/code_review_bot/cli.py#L177.

We should instead add at least a message like events does in case of generic failures: https://github.com/mozilla/code-review/blob/71390ca1bf455fde5db4b34dc271307edf5bb955/events/code_review_events/workflow.py#L203.

abpostelnicu:
We are using `requests` in the `backend` module, so track the `requests.exceptions.HTTPError` and update `phabricator`.
This fixes #984.

### DIFF
--- a/bot/code_review_bot/cli.py
+++ b/bot/code_review_bot/cli.py
@@ -28,6 +28,11 @@ from code_review_tools.log import init_logger
 logger = structlog.get_logger(__name__)
 
 
+LANDO_FAILURE_MESSAGE = (
+    "Static analysis and linting did not run due to a generic failure."
+)
+
+
 def parse_cli():
     """
     Setup CLI options parser
@@ -114,12 +119,17 @@ def main():
 
     # lando the Lando API
     lando_reporting_enabled = "lando" in reporters
+    lando_api = None
+    lando_publish_generic_failure = False
     if lando_reporting_enabled:
         if taskcluster.secrets["LANDO"].get("publish", False):
             lando_api = LandoWarnings(
                 api_url=taskcluster.secrets["LANDO"]["url"],
                 api_key=phabricator["api_key"],
             )
+            lando_publish_generic_failure = taskcluster.secrets["LANDO"][
+                "publish_failure"
+            ]
             reporters["lando"].setup_api(lando_api)
 
     # Load unique revision
@@ -164,7 +174,6 @@ def main():
             w.ingest_autoland(revision)
         else:
             w.run(revision)
-
     except Exception as e:
         # Log errors to papertrail
         logger.error("Static analysis failure", revision=revision, error=e)
@@ -176,9 +185,7 @@ def main():
             extras["error_message"] = str(e)
         w.index(revision, state="error", **extras)
 
-        # Update Harbormaster status
-        w.update_status(revision, state=BuildState.Fail)
-
+        # Update Phabricator
         failure = UnitResult(
             namespace="code-review",
             name="general",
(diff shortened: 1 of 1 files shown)


bot/code_review_bot/cli.py
```diff
@@ -162,6 +165,30 @@ def main():
...
+            revision.target_phid, BuildState.Fail, unit=[failure]
+        )
+
+        # Update Harbormaster status
+        w.update_status(revision, state=BuildState.Fail)
+
+        # Then raise to mark task as erroneous
+        raise
+
     except Exception as e:
```
marco-c, line 192: The problem in #984 is that when an exception happens from anywhere in the bot, we report a build error but we don't say anything to the user.
Instead of handling HTTP errors specifically, let's just report the generic failure whenever there is any exception.

abpostelnicu: OK, I see now, this makes sense.

bot/code_review_bot/cli.py
```diff
@@ -162,6 +164,7 @@ def main():
             w.ingest_autoland(revision)
         else:
             w.run(revision)
+
```
marco-c, line 167: Nit: please remove this extraneous new line

bot/code_review_bot/cli.py
```diff
@@ -176,6 +179,19 @@ def main():
         # Update Harbormaster status
         w.update_status(revision, state=BuildState.Fail)
 
+        failure = UnitResult(
```
marco-c, line 182: Could you report this to lando too, like we do in events?

bot/code_review_bot/cli.py
```diff
@@ -176,6 +179,19 @@ def main():
         # Update Harbormaster status
         w.update_status(revision, state=BuildState.Fail)
```
marco-c, line 180: I don't know if this is still necessary after we add `w.phabricator.update_build_target(`, maybe not

abpostelnicu: I don't think we need it.

bot/code_review_bot/cli.py
abpostelnicu: Sure thing, same error as a generic failure.

bot/code_review_bot/cli.py
```diff
@@ -173,8 +181,30 @@ def main():
...
+            duration=0,
+        )
+
+        w.phabricator.update_build_target(
+            revision.build_target_phid, BuildState.Fail, unit=[failure]
+        )
+
+        # Also update lando
+        if lando_api is not None:
+            try:
```
marco-c, line 200: Let's put it behind the same config as we have in events, so we can disable it if it gets too noisy for develoeprs

abpostelnicu: Sure!

bot/code_review_bot/cli.py
```diff
@@ -173,8 +185,30 @@ def main():
...
+        )
+
+        w.phabricator.update_build_target(
+            revision.build_target_phid, BuildState.Fail, unit=[failure]
+        )
+
+        # Also update lando
+        if lando_publish_generic_failure:
+            try:
+                lando_api.add_warning(
```
marco-c, line 205: We should first delete the pre-existing "in progress" warnings, right?

marco-c (approved):

PR #987, opened 2021-09-23
bot: do not throw exceptions when backend fails.

### LINKED ISSUE #983: Ignore exceptions from the backend
Currently a backend exception can cause a total failure of the bot (e.g. it happened in https://phabricator.services.mozilla.com/D126393).
We should instead ignore backend errors, as they are not really blocking the bot from working.

https://github.com/mozilla/code-review/blob/71390ca1bf455fde5db4b34dc271307edf5bb955/bot/code_review_bot/backend.py#L135 raises, goes back to https://github.com/mozilla/code-review/blob/71390ca1bf455fde5db4b34dc271307edf5bb955/bot/code_review_bot/workflow.py#L290, goes back to https://github.com/mozilla/code-review/blob/71390ca1bf455fde5db4b34dc271307edf5bb955/bot/code_review_bot/workflow.py#L96, exception caught in https://github.com/mozilla/code-review/blob/71390ca1bf455fde5db4b34dc271307edf5bb955/bot/code_review_bot/cli.py#L165

abpostelnicu:
This fixes #983

### DIFF
--- a/bot/code_review_bot/backend.py
+++ b/bot/code_review_bot/backend.py
@@ -61,6 +61,10 @@ class BackendAPI(object):
         }
         backend_revision = self.create("/v1/revision/", data)
 
+        # If we are dealing with None `backend_revision` bail out
+        if backend_revision is None:
+            return
+
         # Create diff attached to revision on backend
         data = {
             "id": revision.diff_id,
@@ -70,6 +74,11 @@ class BackendAPI(object):
             "repository": revision.repository,
         }
         backend_diff = self.create(backend_revision["diffs_url"], data)
+
+        # If we are dealing with a None `backend_revision` bail out
+        if backend_diff is None:
+            revision.issues_url = None
+            return
 
         # Store the issues url on the revision
         revision.issues_url = backend_diff["issues_url"]
@@ -82,7 +91,12 @@ class BackendAPI(object):
             logger.warn("Skipping issues publication on backend")
             return
 
-        assert revision.issues_url is not None, "Missing issues_url on revision"
+        if revision.issues_url is None:
+            logger.warn(
+                "Missing issues_url on revision",
+            )
+            return
+
         published = 0
         for issue in issues:
             payload = issue.as_dict()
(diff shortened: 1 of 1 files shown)


bot/code_review_bot/backend.py
```diff
@@ -131,8 +142,9 @@ def create(self, url_path, data):
         url_post = urllib.parse.urljoin(self.url, url_path)
         response = requests.post(url_post, json=data, auth=auth)
         if not response.ok:
-            logger.warn("Backend rejected the payload: {}".format(response.content))
-        response.raise_for_status()
+            logger.warning("Backend rejected the payload: {}".format(response.content))
+            # return an empty dict
+            return {}
```
marco-c, line 147: We should return `None` instead so it's clearer for callers that there was nothing created, and so that we skip this: https://github.com/mozilla/code-review/blob/9df3342394e283213f6598100e4607c1a934d2cf/bot/code_review_bot/__init__.py#L142

bot/code_review_bot/backend.py
```diff
@@ -131,8 +142,9 @@ def create(self, url_path, data):
         url_post = urllib.parse.urljoin(self.url, url_path)
         response = requests.post(url_post, json=data, auth=auth)
         if not response.ok:
-            logger.warn("Backend rejected the payload: {}".format(response.content))
-        response.raise_for_status()
+            logger.warning("Backend rejected the payload: {}".format(response.content))
```
marco-c, line 145:
```suggestion
            logger.warn("Backend rejected the payload: {}".format(response.content))
```

bot/code_review_bot/backend.py
abpostelnicu: Sure thing!

bot/code_review_bot/backend.py
```diff
@@ -132,7 +143,8 @@ def create(self, url_path, data):
         response = requests.post(url_post, json=data, auth=auth)
         if not response.ok:
             logger.warn("Backend rejected the payload: {}".format(response.content))
-        response.raise_for_status()
+            # return an empty dict
```
marco-c, line 146: The comment no longer applies :P

marco-c (dismissed):
Could you write a test to ensure that, when the backend fails, we are still able to publish?

If you prefer to do it in a follow-up, please file an issue so we don't forget.

abpostelnicu:
> Could you write a test to ensure that, when the backend fails, we are still able to publish?
> 
> If you prefer to do it in a follow-up, please file an issue so we don't forget.

No, I would prefer to do it now, to know that we have this covered!

bot/tests/test_backend.py
```diff
@@ -204,3 +204,34 @@ def test_publication_failures(
             "fix": None,
         }
     ]
+
+
+def test_publication_failures_total(
```
marco-c, line 209: We should also add a test which runs the `run` function of the workflow and, even if the backend fails, it publishes some issues

abpostelnicu:
@marco-c I wanna do this in this release wdyt if we add the tests as a followup?

marco-c:
> @marco-c I wanna do this in this release wdyt if we add the tests as a followup?

WFM, can you file an issue for that?

marco-c (approved):
LGTM, please file a follow-up issue about adding tests.

PR #1045, opened 2021-11-24
bot: make sure we are sending utf-8 when publishing improvement patches.

abpostelnicu:
This should fixes issues like [this](https://phabricator.services.mozilla.com/D132030).


bot/code_review_bot/revisions.py
```diff
@@ -61,7 +61,7 @@ def publish(self, days_ttl=30):
         ), "Only publish on online Taskcluster tasks"
         self.url = taskcluster.upload_artifact(
             "public/patch/{}".format(self.name),
-            self.content,
+            str(self.content, "utf-8"),
```
marco-c, line 64: This is going to do the decoding, we should do the opposite (`self.content` is a string and we want bytes)

PR #1201, opened 2022-06-30
Drop neutrino dependency for the webpack build

vrigal:


vrigal:
Closed in favor of #1211

PR #1202, opened 2022-06-30
Bump Django to 4.0.5

### LINKED ISSUE #1195: Django upgrade to 4.0
Current version (3.1) does not receive any security support, so we need to update to latest stable (4.0).

Usually it's easier to do version by version upgrade (3.2, 4.0). It could be done in a single merge request with one commit per version.

EvaBardou:
Closes #1195   

@marco-c A new setting must be provided with Django 4, I added the logic to support assigning it through the Taskcluster secret like it's done for CORS allowed domains. You'll have to define a new attribute called `csrf-trusted-origins` in the secret with a list of trusted origins (so likely containing the code review backend URL + code review frontend URL). For more information, see [Django official documentation](https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS).

I also had to update the `check` attribute from `code_review_backend.issues.Issue` to `analyzer_check` because of a `SystemCheckError` from Django:
```
django.core.management.base.SystemCheckError: SystemCheckError: System check identified some issues:

ERRORS:
issues.Issue: (models.E020) The 'Issue.check()' class method is currently overridden by <django.db.models.query_utils.DeferredAttribute object at 0x7f9e985b9eb0>.
```
The change was required since there is a function with the same name in the `django.db.models.Model` class, see https://github.com/django/django/blob/main/django/db/models/base.py#L1524.


backend/code_review_backend/issues/serializers.py
```diff
@@ -183,7 +183,7 @@ class IssueCheckStatsSerializer(serializers.Serializer):
 
     repository = serializers.CharField(source="diff__revision__repository__slug")
     analyzer = serializers.CharField()
-    check = serializers.CharField()
+    analyzer_check = serializers.CharField()
```
La0, line 186: By doing that, you change the API output and break the frontend: you should maintain `check` as name and specify the field source instead coming from `analyzer_check`

backend/code_review_backend/issues/migrations/0005_rename_check_issue_analyzer_check.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ("issues", "0004_issue_in_patch"),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name="issue",
+            old_name="check",
+            new_name="analyzer_check",
```
La0, line 17: Why is this change needed ? Is it required by the upgrade ?

La0 (changes requested):

backend/code_review_backend/issues/migrations/0005_rename_check_issue_analyzer_check.py
EvaBardou: Yes it is, there is a function with the same name in the `django.db.models.Model` class, see https://github.com/django/django/blob/main/django/db/models/base.py#L1524. An error is raised by `runserver` if this change isn't applied:
```
django.core.management.base.SystemCheckError: SystemCheckError: System check identified some issues:

ERRORS:
issues.Issue: (models.E020) The 'Issue.check()' class method is currently overridden by <django.db.models.query_utils.DeferredAttribute object at 0x7f9e985b9eb0>.
```

La0: Okay gotcha, please mention it in the description :+1:

abpostelnicu:
This created the following regression https://paste.mozilla.org/QTDOtwmF

abpostelnicu:
In the end it proved to be no regression, thank you @EvaBardou  for the help!

PR #1243, opened 2022-07-28
Bug 1610715 - Add a hint when error comes from a parent revision

### LINKED ISSUE #553: Add a comment when errors are published outside of patch

vrigal:
Follow-up of #561 and #690
Closes #553

### DIFF
--- a/bot/code_review_bot/report/phabricator.py
+++ b/bot/code_review_bot/report/phabricator.py
@@ -248,7 +248,9 @@ class PhabricatorReporter(Reporter):
 
         # Add extra hint when errors are published outside of the patch
         defects_details = ""
-        external_failures_count = sum(
+        # Only display the hint when the revision has parents in his stack
+        rev_parents_count = len(self.api.load_parents(revision.phid))
+        external_failures_count = rev_parents_count and sum(
             1
             for issue in issues
             if issue.is_publishable() and not revision.contains(issue)
--- a/bot/code_review_bot/report/phabricator.py
+++ b/bot/code_review_bot/report/phabricator.py
@@ -248,13 +248,15 @@ class PhabricatorReporter(Reporter):
 
         # Add extra hint when errors are published outside of the patch
         defects_details = ""
-        if any(
-            issue.is_publishable() and not revision.contains(issue) for issue in issues
-        ):
-            if nb == 1:
-                defects_details = " (in a parent revision)"
-            else:
-                defects_details = " (some in a parent revision)"
+        external_failures_count = sum(
+            1
+            for issue in issues
+            if issue.is_publishable() and not revision.contains(issue)
+        )
+        if nb == 1 and external_failures_count == 1:
+            defects_details = " (in a parent revision)"
+        elif nb > 1 and external_failures_count > 0:
+            defects_details = f" ({external_failures_count} in a parent revision)"
 
         if nb > 0:
             comment = COMMENT_FAILURE.format(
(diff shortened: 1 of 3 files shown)


bot/code_review_bot/report/phabricator.py
```diff
@@ -246,9 +246,20 @@ def pluralize(word, nb):
...
 
+        # Add extra hint when errors are published outside of the patch
+        defects_details = ""
+        if any(
+            issue.is_publishable() and not revision.contains(issue) for issue in issues
+        ):
+            if nb == 1:
+                defects_details = " (in a parent revision)"
+            else:
+                defects_details = " (some in a parent revision)"
```
marco-c, line 257: Maybe mention the actual number instead of "some"?

vrigal:
I have not tested this yet.

@marco-c this thread is a bit confusing to me : https://github.com/mozilla/code-review/pull/561/files#r434249237
I'm not sure if it is possible to list revision's parents easily.
Here is a sample of the loaded revision (from test data):
```
{
  'id': 51,
  'type': 'DREV',
  'phid': 'PHID-DREV-azcDeadbeef',
  'fields': {
    'title': 'Static Analysis tests',
    'authorPHID': 'PHID-USER-crasher',
    'status': {'value': 'needs-review', 'name': 'Needs Review', 'closed': False, 'color.ansi': 'magenta'},
    'repositoryPHID': 'PHID-REPO-coffeee',
    'diffPHID': 'PHID-DIFF-test',
    'summary': 'Nice summary',
    'dateCreated': 1512389294,
    'dateModified': 1512401613,
    'policy': {'view': 'public', 'edit': 'users'},
    'bugzilla.bug-id': '1234567'},
  'attachments': {}
}
```
I can also access a list of commits for the diff with a `parents` fields listing commit IDs, but I think it won't be helpful.

bot/code_review_bot/report/phabricator.py
```diff
@@ -246,9 +246,20 @@ def pluralize(word, nb):
         # Build top comment
         nb = len(issues)
 
+        # Add extra hint when errors are published outside of the patch
+        defects_details = ""
+        if any(
+            issue.is_publishable() and not revision.contains(issue) for issue in issues
+        ):
```
marco-c, line 253: We should assert that there is a parent revision here, otherwise it might be confusing if there isn't one (in theory there should be, otherwise where does the issue come from? But you never know, maybe we are missing something...).

See also https://github.com/mozilla/code-review/pull/561#discussion_r434249237.

vrigal: I saw this thread and commented here https://github.com/mozilla/code-review/pull/1243#issuecomment-1197939329.
I don't know much Phabricator API, but I'll look for a way to get this information later on.

marco-c: Yeah when I was writing my comment I hadn't seen your comment yet :)

You can use this function for getting parent information: https://github.com/mozilla/libmozdata/blob/e8c19af0552dfffdff4bfcb69d423a73a3d9f280/libmozdata/phabricator.py#L345.

abpostelnicu (approved):

PR #1253, opened 2022-08-02
Use sentry_sdk instead of raven and improve Sentry grouping

### LINKED ISSUE #1225: Improve Sentry grouping
Sentry for code-review and code-coverage is almost unusable because we have too many errors which are not grouped together.

E.g. we have one `Mercurial error on diff (error='patching file XXX` for each possible file, while we should have only one `Mercurial error on diff` that groups all of them together. 

We should either find a way to make Sentry group all of those issues together, or modify the log messages to avoid including specific information about errors.

Here's where that example warning comes from: https://github.com/mozilla/libmozevent/blob/f1d8c03c418bebe45f468f84dabfb85d4d533685/libmozevent/mercurial.py#L386.

EvaBardou:
Closes #1225 

Please test carefully this PR, I am not really familiar with **Sentry** and **Papertrail** and may have missed some details that may no longer appear in the logs or in Sentry issues and are very crucial for you. As you will see in the following screenshots, I was able to test the **Sentry** integration using a real instance. Regarding **Papertrail**, I simply redirected the output to `/dev/log`, rather than to a real Papertrail instance using a `host` and a `port`, so that I could check the format of the logs but that's all.

## The issue

As stated in issue #1225, the concern encountered is that the Sentry issue name almost always includes a unique context which prevents issues from being grouped together by having several different events attached.

**Currently we have:**
Issue number 1 (`context_element=1`) => 1 attached event
Issue number 1 (`context_element=2`) => 1 attached event
Issue number 2 (`other_context_element=1`) => 1 attached event

**The requested modifications would allow us to obtain the following grouping:**
Issue number 1 => 2 attached events, event 1 (`context_element=1`) & event 2 (`context_element=2`)
Issue number 2 => 1 attached event

## The illustrated results

See the following screenshots with test data that illustrate the new grouping.

**Before:**
_Two issues with one attached event_
Two issues with one attached event

**After:**
_One issue with two attached events_
One issue with two attached events

_Additional Data section for the first event_
Additional Data section for the first event

## The implementation

I made a mix of the two solutions suggested by Bastien in [this comment](https://github.com/mozilla/code-review/issues/1225#issuecomment-1202164904), meaning, I kept `structlog` to avoid having to modify a lot of lines and also having to submit a PR on `libmozevent` and I didn't add the new dependency called `structlog_sentry` that we mentioned.

Each log follows a specific path:
1. The log is created by using a `logger.info`, `logger.warning`, etc, knowing that the logger is the `structlog` one.
2. It passes through several `structlog` processors as before, except that a new processor can also transpose the remaining attributes of the `event_dict` of this log into `extra` for `logging.log`.
3. 
      - The log is outputted on `stdout` using a custom format (I tried to reproduce the same structure as the one from `logbook`, minus the colorization)
      - If **Papertrail** is activated, a handler of type `SysLogHandler` is in charge of transmitting the log using another custom format
      - If **Sentry** is activated, thanks to the logging integration of `sentry_sdk`, the log is captured and sent back to Sentry. `sentry_sdk` will take care of getting the `extra` attribute of the log to translate it into **Additional Data** to be added on each event.

The problem with the previous implementation was that [one of the processors](https://github.com/mozilla/code-review/blob/master/tools/code_review_tools/log.py#L15) was adding the context in the title of each issue rather than serving it as **Additional Data** as it does now.

I was able to replace the `raven` dependency with the `sentry_sdk` one and drop the use of `logbook` completely.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,6 +1,6 @@
 [settings]
 known_first_party = code_review_backend,code_review_bot,code_review_tools,code_review_events,conftest
-known_third_party = dj_database_url,django,drf_yasg,influxdb,jsone,jsonschema,libmozdata,libmozevent,logbook,parsepatch,pytest,raven,requests,responses,rest_framework,rs_parsepatch,sentry_sdk,setuptools,structlog,taskcluster,yaml
+known_third_party = dj_database_url,django,drf_yasg,influxdb,jsone,jsonschema,libmozdata,libmozevent,parsepatch,pkg_resources,pytest,requests,responses,rest_framework,rs_parsepatch,sentry_sdk,setuptools,structlog,taskcluster,yaml
 force_single_line = True
 default_section=FIRSTPARTY
 line_length=159
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -6,6 +6,5 @@ djangorestframework==3.13.1
 drf-yasg==1.21.3
 gunicorn==20.1.0
 psycopg2-binary==2.9.3
-sentry-sdk==1.7.2
 sqlparse==0.4.2
 whitenoise==6.2.0
--- a/tools/code_review_tools/log.py
+++ b/tools/code_review_tools/log.py
@@ -3,29 +3,28 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import logging
+import logging.handlers
 import os
+import sys
 
-import logbook
-import logbook.more
-import raven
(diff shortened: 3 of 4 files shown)


La0 (changes requested):
LGTM code-wise, but you need to explain the new setup in a comment (especially how `extra` are managed by logging + sentry) and make a nice summary of what is going on in the description

La0 (approved):
Thanks for the comments & description !

marco-c (approved):
Mostly a rubber stamp as I know nothing about these libraries :P
We'll closely watch the effects of this after landing!

PR #1276, opened 2022-08-16
Fix "Attempt to overwrite '*' in LogRecord"

EvaBardou:
Fix a bug that appeared after merging #1253, that raises an error `Attempt to overwrite '*' in LogRecord` (with `*`, a common attribute from `LogRecord`) when several logs were processed by `structlog` processors.

In `logging` documentation:
> The keys in the dictionary passed in _extra_ should not clash with the keys used by the logging system. (See the section on [LogRecord attributes](https://docs.python.org/3/library/logging.html#logrecord-attributes) for more information on which keys are used by the logging system.)

The line raising the error in `logging`:
https://hg.python.org/cpython/file/3.5/Lib/logging/__init__.py#l1394

The quick fix implemented in this PR is to rename all clashing attributes by adding a `_` character.

### DIFF
--- a/tools/code_review_tools/log.py
+++ b/tools/code_review_tools/log.py
@@ -84,6 +84,26 @@ def setup_sentry(name, channel, dsn):
         sentry_sdk.set_context("task", {"task_id": task_id})
 
 
+class RenameAttrsProcessor(structlog.processors.KeyValueRenderer):
+    """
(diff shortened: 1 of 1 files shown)


abpostelnicu (approved):

abpostelnicu:
Thank you! We might be able to do a release after this :)

PR #1296, opened 2022-08-31
Phabricator followup comments

### LINKED ISSUE #1295: Use new_for_revision information in followup Phabricator comments
The bot should now use the `new_for_revision` information provided by the backend, changing the way we publish followup comments (on diffs posted after the first one.)

The comment on the initial diff should stay as is (either silent when there are no issues, or report all found issues).

Followup comments should now be:
- skipped when nothing changed (no new issues, no issues resolved)
- updated otherwise:
   - report new issues as currently (adding a unit result to mention them in the diff)
   - only mention previous unresolved issues (no unit results for these)
   - mention resolved issues (issues that were present in older diffs that are now gone)
   
The last point may require an extra API request to get the list of issues from the previous diff known from the backend.

vrigal:
Closes #1295


vrigal:
Example compared to the current code review bot: https://phabricator.services.mozilla.com/D156648#5159188

vrigal:
If #1318 is merged and we add a DB constraint (hashes unicity), the code could be simplified a bit (no need of `defaultdict` struct)

PR #1398, opened 2022-11-08
bot: add force publish field

abpostelnicu:
Some checks, like some clang-tidy specific ones, they do not assert errors an the places where they assert the warnings are not part of the patch. These checks are for code quality improvement over time and the developer should still need to see a warning maybe he wants to contribute. 
This is why we use this extra field.


marco-c (approved):
LGTM!

PR #1449, opened 2022-12-05
Bug 1803963 - Prevent the horizontal delimiter to act as header in Phabricator comments

vrigal:
The text was simply appearing as a title because the delimiter (`---`) was directly placed under that block.

Reformatting the `COMMENT_DIFF_FOLLOWUP` variable (with an extra line break at the end) was enough, but I also updated the horizontal rule to prevent that behavior in the future (supplementary line breaks will not be rendered as markdown).

bugzilla_image


marco-c (approved):
LGTM, thanks!

PR #1462, opened 2022-12-09
Enhance Issue model to support multiples diffs/revisions

### LINKED ISSUE #1456: Enhance Issue storage to support lots of duplicates
### Current state

```mermaid
graph BT

    Issue --> Diff
    Diff  --> Revision
    Revision --> Repository

```

The hierarchy is simply defined using foreign keys between models (a revision belongs to a single repo, a diff belongs to a single revision, ...)

The issue with this approach appears when we have a lot of duplication of issues: we are forced to store multiple times the same data simply to represent a link between a diff/revision/repository and an issue. We are already seeing this on large revisions with multiple diffs, but this is non-viable as for ingesting mozilla-central huge logs of issues (see #1436). 

### Proposal

```mermaid
graph BT

    Revision --> Repository
    Diff --> Revision
    IssueDiff --> Diff
    IssueDiff --> Issue
```

We would introduce a new model named `IssueDiff` that links to both an `Issue` and a `Diff`. This replaces the 1-N relationship between both models by a many-to-many relation.

Unique database constraints would be created on:
- Issue `hash + analyzer + file` so that we only represent a unique issue per analyzer and file
- IssueDiff `issue + diff` so that we only have the an issue once per diff

This would not change the current usage at all, but decrease the space required to store all these new issues from MC. The query performance should also be improved both in write and reads. 
- writes should be in average faster as pre-existing issues only require to add an entry in `IssueLink` to reference them
- reads should be faster overall as the database should scan a lot less rows 

Data migration would be easy to implement: all current links would be converted to `IssueDiff` links. There is probably no duplicates on issue hash (but we would need to check against a production database dump).

The existing Rest API should not change, we simply need to update some queryset and serializer to access the `Diff` instances through this new relation.

vrigal:
Closes #1456

---
## How to run the migration

The migration is done in 2 parts. The first one creates the new M2M table (and update models references).
The second one copy all the existing links between Issue and Diff to the new table, then drops the FK.
This migration is long (tens of hours on the production DB), but non blocking:
* New detected issue will still be added to the backend
* The frontend will be accessible but will show recent issues until the migration is fully finished.

note1: If we notice that DB keep locked between each iteration, it is possible to set smaller chunks (default to 1000).
note2: Running that migration against a copy of the production database took a few hours (estimated <10h).

---
## Features
* Non reversible migration from the Diff FK to the M2M table
* Update API endpoints (most references to an issue's repository uses the link to a revision)
  * (IssueCheckDetails](https://api.code-review.moz.tools/docs/#operations-check-check_read) `diff` as been replaced to `diffs`, listing multiple diffs with their repository. This view is used by the frontend to display statistics.
* Updated the `load_issues` management command (somehow does not seems to work - repository lookup fails and getArtifact does not allow redirect)
* I updated the cleanup management command. Now old revisions are removed but issues may still be linked to a recent revision.

### DIFF
--- a/backend/code_review_backend/issues/migrations/0007_issuelink_swap.py
+++ b/backend/code_review_backend/issues/migrations/0007_issuelink_swap.py
@@ -9,7 +9,7 @@ from math import ceil
 
 from django.db import migrations
 
-ISSUES_INSERT_SIZE = 5000
+ISSUES_INSERT_SIZE = 1000
 
 
 def generate_issue_links(apps, schema_editor):
@@ -40,8 +40,9 @@ def generate_issue_links(apps, schema_editor):
     slices = ceil(issues_count / ISSUES_INSERT_SIZE)
 
     for index in range(0, slices):
+        current = min((index + 1) * ISSUES_INSERT_SIZE, issues_count)
         print(
-            f"[{index + 1}/{slices}] Initializing Issues references on the M2M table."
+            f"[{current}/{issues_count}] Initializing Issues references on the M2M table."
         )
         issues = qs[index * ISSUES_INSERT_SIZE : (index + 1) * ISSUES_INSERT_SIZE]
         IssueLink.objects.bulk_create(
@@ -69,6 +70,7 @@ class Migration(migrations.Migration):
             generate_issue_links,
             reverse_code=None,
             elidable=True,
+            atomic=False,
         ),
         # Drop old FK
         migrations.RemoveField(


vrigal:
You can follow those steps to generate a lot of issues on the backend and run the migration:
* Checkout to `vrigal:fix-load-issues-command`, make sur your schema is up-to-date (`./manage.py migrate`)
* Ensure you have at least a repository with `{"slug": "mozilla-central", "url": "https://hg.mozilla.org/mozilla-unified"}`. You can create them from the admin interface otherwise.
* Run the command `./manage.py load_issues`. It should generate tens of thousands issues in a few minutes. You can resume that command anytime.
* Checkout to this branch (`vrigal:issue-link`) and perform migration (`manage.py migrate issues`).

backend/code_review_backend/issues/serializers.py
```diff
@@ -170,24 +170,28 @@ class IssueCheckSerializer(IssueSerializer):
...
+        fields = IssueSerializer.Meta.fields + ("diffs",)
 
 
 class IssueCheckStatsSerializer(serializers.Serializer):
     """
     Serialize the usage statistics for each check encountered
     """
 
-    repository = serializers.CharField(source="diff__revision__repository__slug")
+    repository = serializers.SlugField(source="revisions__repository__slug")
```
La0, line 185: How does that work when an issue is in multiple revisions with multiple repos ?

La0:
Just delete is good. 

Per the comment on the [nullable FK](https://github.com/mozilla/code-review/blob/master/backend/code_review_backend/issues/models.py#L84) we will actually support issues around the whole tree, and could even make this FK non nullable at a later stage

vrigal:
> 

Alright, there are no issues like this on prod so it should be fine !

La0:
So you could make it non-nullable in the migration then

vrigal:
Is it possible to add a constraint on the M2M depending on issues ? \
And then how could we add an issue without it link entry ?

vrigal:
I ran the migration on the shared prod database dump. It was split into 4773 chunks and took 1+ hour.

I will split the migration into two (create the table, then populate it), so there is no table lock on the M2M during the data copy. The frontend will not show most of the issue during this time but new issues could still be published from the bot.

We could try a `INSERT INTO SELECT` pattern for better performance, filtering issues with no diff (still not migrated) with a specific isolation level to allow inserts.

vrigal:
Splitting the migration into two seems to work. I was able to read and insert data (btw this does not alter the select query which excludes issues with no diff), but it took 15 hours to complete (executed from didyme to xenarque) .
```
real    894m27.891s
user    14m1.096s
sys     0m5.369s
```

vrigal:
As expected, using `INSERT INTO … SELECT` is much faster, but locks `issue_issue_link` for writting. Also handling the auto increment would be quite complex with concurrency.
It is a heavy operation too (stopping the migration was difficult, I had to use `pg_cancel_backend`). 

It took <at least 20min> to complete (I will edit once finished).
I pushed to another branch anyway `vrigal:issue-link-insert-select`.

La0:
Ok, we'll go forward with the split migration. Worst case, we'll run the `INSERT INTO SELECT` query manually on the db server.

vrigal:
I'm testing 048ece6f3eaabff7f8f87188f10d1cc9e64143ef, it seems to work fine.
I can create issues (including via the API) and they are created between two bulk inserts.
```python
>>> # Started the migration
>>> Issue.objects.count()
23860577
>>> IssueLink.objects.count()
0
>>> issue1 = Issue.objects.create(hash="test1")
>>> issue_link1 = IssueLink.objects.create(issue=issue1, diff=Diff.objects.first(), revision=Revision.objects.first())
>>> issue1.id
UUID('3e4d420d-c6b0-4d84-b024-bcc79d5df1a5')
>>> issue_link1.id
390001
>>> issue2 = Issue.objects.create(hash="test2")            
>>> issue_link2 = IssueLink.objects.create(issue=issue2, diff=Diff.objects.first(), revision=Revision.objects.first()
)
>>> issue_link2.id                                         
525002
>>> issue_link2.diff_id
180406
>>> # Creating an issue via the API
>>> Issue.objects.get(id="77106fd5-affb-42b5-a94d-465aeb0cfc7b").issue_links.get().id
2475003
```

vrigal:
I noticed the last step of the migration was starving thus locking the database for a few hours.
I updated the second migration so it copy the data in multiple transactions. It also allows to stop the migration and continue it later. Furthermore, it should probably not conflict anymore with the `cleanup_issues` management command, as issues are copied from the most recent to the oldest (the migration should succeed after trying to copy empty slices of data).

La0:
Much better time :100: 

We'll be able to fully test your changes on our local dump (post migration) then

vrigal:
Finally it took around 10 hours. I don't know why, but the first few chunks are copied much faster than the last ones. \
We should probably reduce chunks size (e.g. 1000 instead of 5000) to be sure we do not lock the database for a long period.

@marco-c could you take a look at this migration please ? Most of the other things I am working on depends on this work.

La0 (approved):
DB Fully restored then cleaned up to keep last year issues (370k / 23 millions).

Migration took ~ 2 minutes :tada:.

:+1: for me, we should try this on testing

PR #1476, opened 2022-12-13
Fix load issues management command

vrigal:
Requires https://github.com/mozilla/code-review/pull/1462 (merged)

### DIFF
--- a/backend/code_review_backend/issues/management/commands/load_issues.py
+++ b/backend/code_review_backend/issues/management/commands/load_issues.py
@@ -175,7 +175,7 @@ class Command(BaseCommand):
 
     def build_revision_and_diff(self, data, task_id):
         """Build or retrieve a revision and diff in current repo from report's data"""
-        repository = Repository.objects.get(slug=data["target_repository"])
+        repository = Repository.objects.get(url=data["target_repository"])
         revision, _ = repository.revisions.get_or_create(
             id=data["id"],
             defaults={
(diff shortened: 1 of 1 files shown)


backend/code_review_backend/issues/management/commands/load_issues.py
```diff
@@ -166,7 +172,13 @@ def load_local_reports(self):
 
     def build_revision_and_diff(self, data, task_id):
         """Build or retrieve a revision and diff in current repo from report's data"""
-        repository = Repository.objects.get(slug=data["target_repository"])
+        try:
+            repository = Repository.objects.get(url=data["target_repository"])
+        except Repository.DoesNotExist:
+            logger.warning(
+                "No repository found with URL {data['repository']}, skipping."
```
La0, line 179: This lacks the `f-string` prefix

La0 (changes requested):

La0 (commented):
I ran the script against the prod dump, and I'm getting conflicts issues on Diff (already present issue in DB with the same PHID, but not the same ID / revision link ?)

Can you test it against the same DB please ?

vrigal:
That error happens because a diff was already existing, and has been created from another repository.
From what I understand this is intended behavior at diff's level (https://github.com/mozilla/code-review/pull/273).

I made the change to get_or_create the diff according to its ID only.

marco-c (approved):

PR #1502, opened 2023-01-16
In `build_tak` for tasks that do not match anything returrn None.

abpostelnicu:
This type of tasks doesn't expose issues.json so we should ignore it.

### DIFF
--- a/bot/code_review_bot/workflow.py
+++ b/bot/code_review_bot/workflow.py
@@ -442,7 +442,7 @@ class Workflow(object):
             # Log cleanly on autoland unknown tasks
             logger.info("Skipping unknown task", id=task_id, name=name)
         else:
-            return DefaultTask(task_id, task_status)
+            return None
 
     def update_status(self, revision, state):
         """
--- a/bot/tests/test_remote.py
+++ b/bot/tests/test_remote.py
@@ -351,7 +351,7 @@ def test_unsupported_analyzer(mock_config, mock_revision, mock_workflow, mock_ba
     )
     issues = mock_workflow.run(mock_revision)
     assert len(issues) == 0
-    assert mock_revision._state == BuildState.Fail
+    assert mock_revision._state == BuildState.Pass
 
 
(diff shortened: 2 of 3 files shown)


bot/code_review_bot/workflow.py
```diff
@@ -436,6 +436,8 @@ def build_task(self, task_status):
             return ExternalTidyTask(task_id, task_status)
         elif name == "source-test-taskgraph-diff":
             return TaskGraphDiffTask(task_id, task_status)
+        elif name.startswith("source-test-shadow-scheduler"):
```
marco-c, line 439: If we do this, we have to maintain a list of tasks that we don't care about, I think it's preferable to just return None in the last `else`, since if the artifact is not available we will never be able to parse it in DefaultTask anyway.

bot/tests/test_remote.py
```diff
@@ -351,7 +351,6 @@ def test_unsupported_analyzer(mock_config, mock_revision, mock_workflow, mock_ba
     )
     issues = mock_workflow.run(mock_revision)
     assert len(issues) == 0
-    assert mock_revision._state == BuildState.Fail
```
marco-c, line 354: The assertion should change to something like BuildState.Pass if we now expect this to be an acceptable behavior

bot/tests/test_workflow.py
```diff
@@ -97,11 +96,8 @@ def test_taskcluster_index(mock_config, mock_workflow, mock_try_task):
         ("source-test-clang-format", ClangFormatTask, True),
         ("source-test-taskgraph-diff", TaskGraphDiffTask, False),
         ("source-test-taskgraph-diff", TaskGraphDiffTask, True),
-        ("source-test-unsupported", DefaultTask, False),
-        ("source-test-unsupported", DefaultTask, True),
-        # On autoland, we only support source-test tasks
-        ("totally-unknown", DefaultTask, False),
-        ("totally-unknown", None, True),
```
marco-c, line 104: Let's keep these instead of removing them but just update the expectations

marco-c (approved):

PR #1515, opened 2023-01-17
Bug 1799017 - Migrate code-review from AWS to GCP

MasterWayZ:


marco-c:
@MasterWayZ we might run into the same issue as https://bugzilla.mozilla.org/show_bug.cgi?id=1798886.

marco-c:
Unfortunately not, it's only visible after we make a release and the bot runs on taskcluster

PR #1552, opened 2023-02-13
API endpoint listing issues for a path

### LINKED ISSUE #1540: Build API endpoint to list known issues for a path
This is a part of before/after.

We need a new backend API endpoint to list known issues for a given repository and file path.

The endpoint url should be simply `/issues/<repo_slug>` and support the following query filters:
-  `path`: required parameter, filters the issues for this given path
- `revision`: when set, the endpoint will **first** check if there are any issues for that path + repository + exact revision
   - if not, we fallback to the date filter below
- `date`: when set to a datetime, the endpoint will lookup the closest known revision before that date, going back to 2 days maximum. Once a revision is found, the endpoint returns all the issues for that path and revision.

If no issue is found, the endpoint simply return an empty list (no error).

The endpoint must be paginated as the list of issues can be long for some paths.

vrigal:
Closes #1540

### DIFF
--- a/backend/code_review_backend/issues/api.py
+++ b/backend/code_review_backend/issues/api.py
@@ -3,6 +3,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from collections import defaultdict
 from datetime import date
 from datetime import datetime
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


La0 (approved):
We'll need to know the Phabricator ID of the base revision from the bot, bit there is no other way to identify the base revision in the backend (as there is no storage for the mercurial hash on the rev)

PR #1565, opened 2023-02-27
Support Dockerflow in events

### LINKED ISSUE #1557: Support dockerflow for events
The events web server should support docker flow (using work from https://github.com/mozilla/libmozevent/issues/97)

- The CI should build the version.json file and make it available in the docker image
- When running in development, the endpoint should be disabled (just passing `None` should remove the loading of JSON content).

EvaBardou:
Depends on https://github.com/mozilla/libmozevent/pull/98 being merged + a new release for `libmozevent`

Closes #1557

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -254,7 +254,7 @@ tasks:
                   - /bin/sh
                   - -lxce
                   - "git clone --quiet ${repository} /code-review && cd /code-review && git checkout ${head_rev} -b build
-                    && backend/build.sh ${head_rev} ${head_branch} ${repository} ${channel}"
+                    && events/build.sh ${head_rev} ${head_branch} ${repository} ${channel}"
                 artifacts:
                   public/code-review-events.tar.zst:
                     expires: {$fromNow: '2 weeks'}
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -250,21 +250,11 @@ tasks:
                   privileged: true
                 maxRunTime: 3600
                 image: "${taskboot_image}"
-                env:
-                  GIT_REPOSITORY: ${repository}
-                  GIT_REVISION: ${head_rev}
                 command:
-                  - taskboot
-                  - build
-                  - --image
-                  - mozilla/code-review
-                  - --tag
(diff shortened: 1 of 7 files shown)


.taskcluster.yml
```diff
@@ -250,21 +250,11 @@ tasks:
...
-                  - "${channel}"
-                  - --tag
-                  - "${head_rev}"
-                  - --write
-                  - /events.tar
-                  - events/docker/Dockerfile
+                  - /bin/sh
+                  - -lxce
+                  - "git clone --quiet ${repository} /code-review && cd /code-review && git checkout ${head_rev} -b build
+                    && backend/build.sh ${head_rev} ${head_branch} ${repository} ${channel}"
```
La0, line 257: I guess that's a bad copy-paste: you are building the backend instead of events here.

You need to call `events/build.sh` here.

events/code_review_events/workflow.py
```diff
@@ -375,7 +378,9 @@ def __init__(self, cache_root):
         # Run webserver & pulse on web dyno or single instance
         if not heroku.in_dyno() or heroku.in_web_dyno():
             # Create web server
-            self.webserver = WebServer(QUEUE_WEB_BUILDS)
+            self.webserver = WebServer(
+                QUEUE_WEB_BUILDS, version_path=f"{BASE_DIR}/version.json"
```
La0, line 382: Please use `os.path.join` here

La0 (changes requested):
I built the docker image in taskboot using

```
docker run --privileged -it -v $(pwd):/code-review mozilla/taskboot:0.3.6 sh
```

Then inside the container:

```
./events/build.sh xxxrevyyyzzz master github.com/code-review prod
```

Back on the host, I used `docker cp` to retrieve the `/events.tar.zst` image and load it, then execute it.

Both heartbeat endpoints are working as expected, but the version one does not.

.taskcluster.yml
EvaBardou: oops, good catch

La0 (approved):
The docker image built now provides the version on `/__version__` :+1:

PR #1573, opened 2023-03-01
bot: do not log warning when we don't report anything on a per line basis.

abpostelnicu:
This will resolve warnings like: https://mozilla.sentry.io/issues/3532261274/?alert_rule_id=12607839&alert_timestamp=1677568283919&alert_type=email&environment=production&project=6273985 that are not actionable and the linter still runs without issues.

### DIFF
--- a/bot/code_review_bot/__init__.py
+++ b/bot/code_review_bot/__init__.py
@@ -86,7 +86,7 @@ class Issue(abc.ABC):
         self.line = positive_int("line", line)
         self.nb_lines = positive_int("nb_lines", nb_lines)
 
-        # Support line 0, with a Sentry warning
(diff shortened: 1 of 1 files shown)


marco-c:
@abpostelnicu which linter is returning 0 as line?

bot/code_review_bot/__init__.py
```diff
@@ -88,7 +88,7 @@ def __init__(
 
         # Support line 0, with a Sentry warning
```
marco-c, line 89: Can you adjust the comment to explain that we get line 0 for full-file issues? And remove the part "with a Sentry warning"

marco-c (approved):

PR #1578, opened 2023-03-02
Support empty chunk during issues bulk creation in the backend

vrigal:
Failure found ingesting `MOZILLA_CENTRAL_TASK_GROUP_ID="DzURrYLmTSqJqdCofMXa8w"`


marco-c (approved):

marco-c:
@vrigal how do we reach this situation? Is there a chunk in which all issues are missing hash? Can you check if there is a pattern?

vrigal:
Sorry for the late reply @marco-c.
This is indeed a chunk in which all issues are missing hash.
This is due to that particular decision task, because a lot of successive issues (> than 100) have an invalid file path starting with `../../../../mobile/android/geckoview/src/main/java/org/mozilla/`.
I never reproduced that case otherwise.

marco-c:
I wonder if this is a problem in Android linting tasks, maybe we should fix it at the source. @abpostelnicu could you take a look?

vrigal:
Reproduced with revision 175516 (`MOZILLA_CENTRAL_TASK_GROUP_ID="R_GdFCC5RT-71o7L0RBVZQ"`)

vrigal:
@marco-c Could you fix the Android linting task ? Because that behavior would break before/after as only a part of the issues are published on those revisions.

bot/code_review_bot/backend.py
```diff
@@ -121,6 +121,9 @@ def publish_issues(self, issues, revision, mercurial_repository=None, bulk=0):
                         continue
                     valid_data.append((issue, issue.as_dict(issue_hash=issue_hash)))
 
+                if not valid_data:
+                    continue
```
marco-c, line 125: Let's add a warning here, I know we have one for "Missing issue hash", but this would be a little different (warning when no issue was valid).

marco-c (approved):

marco-c (approved):

Issue #1425, opened 2022-11-24
Add API endpoint to list all known issues from a given repository and file path


La0:
A new endpoint should be exposed as `/repository/<slug>/issues/<path>` to allow listing all known issues for a given repository.

The issues list would be paginated, and sorted by creation time.

As the issues would be regularly cleaned up, we should only retrieve (semi) recent issues.

Some additional filters could be extended:
- a revision date filter, to list issues that happened before a given revision date
- a analyzer / check filter, to list issues produced by a given check (or analyze)

marco-c:
Given that for a given Phabricator revision we know its parent revision, we could use the revision instead of the date (i.e. make the API retrieve "all issues that exist at revision X").

vrigal:
This is probably a duplicate of #1540 (closed with #1552)

PR #1579, opened 2023-03-03
Filter by date revision on the target repository listing issues

vrigal:
Follow up of #1425

Otherwise the most recent revision may be from any other repository from the required one, resulting in an empty queryset (ie. `qs.filter(revisions__id=<recent_rev>, revisions__repository__slug=<other_repo>)`).


vrigal:
I noticed that revisions created ingesting Try pushes always are set with the `mozilla-central` repository.
No revision on prod are link on the try repository, making it impossible to differentiate issues detected from try or mozilla-central.
I'll open an issue and work on this.

vrigal:
Done in #1581

PR #1581, opened 2023-03-03
Compare issues with known ones in the backend

vrigal:


vrigal:
The first part (tagging new issues) is complete whereas the second part (Phabricator publication) is WIP.

Comparing issues between successive diffs is complex because the backend does not know which issues are "new".

bot/code_review_bot/report/phabricator.py
```diff
@@ -22,7 +23,11 @@
 BUG_REPORT_URL = "https://bugzilla.mozilla.org/enter_bug.cgi?product=Developer+Infrastructure&component=Source+Code+Analysis&short_desc=[Automated+review]+THIS+IS+A+PLACEHOLDER&comment=**Phabricator+URL:**+https://phabricator.services.mozilla.com/...&format=__default__"
 
 COMMENT_FAILURE = """
-Code analysis found {defects_total}{defects_details} in the diff [{diff_id}]({phabricator_diff_url}):
+Code analysis found {defects_total} in diff [{diff_id}]({phabricator_diff_url}):
+"""
+
+COMMENT_FAILURE_OUTSIDE = """
+Code analysis found {defects_total} outside of diff [{diff_id}]({phabricator_diff_url}):
```
marco-c, line 30: Instead of splitting in two separate sentences, we could have a single one with the total number of defects (in the end they have the same level of importance, it doesn't matter if they are in the lines modified by the patch or not, it only matters if they were introduced by the patch):
```
Code analysis found {defects_total} introduced by diff [{diff_id}]({phabricator_diff_url}):
```

vrigal:
Actually there seems to be very few issues ingested from mozilla-central on production (https://api.code-review.moz.tools/v1/issues/mozilla-central/), which will cause that implementation to report each issue as new.

The API endpoints falls back to a revision in the last two days (in case the base revision do not exist for mozilla central), and no issues are actually found in that case : https://api.code-review.moz.tools/v1/issues/mozilla-central/?date=2023-04-14.

bot/code_review_bot/__init__.py
```diff
@@ -135,21 +139,31 @@ def is_publishable(self):
...
             return self.on_backend["publishable"]
 
         # Fallback to in_patch detection
+        return self.is_in_patch()
+
+    def is_in_patch(self):
         return self.revision.contains(self)
 
-    def build_hash(self, local_repository=None):
+    def get_hash(self, local_repository=None):
```
La0, line 156: It would be nicer to use [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) and moving `local_repository` to a setting

bot/code_review_bot/__init__.py
```diff
@@ -135,21 +139,31 @@ def is_publishable(self):
...
+            return True
+
         # Then check if the backend marks this issue as publishable
         if self.on_backend is not None:
             return self.on_backend["publishable"]
 
         # Fallback to in_patch detection
+        return self.is_in_patch()
+
+    def is_in_patch(self):
```
La0, line 153: Set as a property `in_patch`

bot/code_review_bot/revisions.py
```diff
@@ -133,6 +134,15 @@ def namespaces(self):
             "phabricator.diffphid.{}".format(self.diff_phid),
         ]
 
+    @property
+    def before_after_feature(self):
+        """
+        Randomly run the before/after feature depending on a configured ratio.
+        All the diffs of a revision must be analysed with or without the feature.
+        """
+        random.seed(self.id)
```
La0, line 143: That does not work, the seed is statically set, but different calls to `random()` will generate different numbers (just try it in a python shell).

You need to remove the random component, and just use the last 4 digits as decimals

bot/code_review_bot/workflow.py
```diff
@@ -95,8 +96,36 @@ def run(self, revision):
...
+            if not base_rev_changeset:
+                logger.info(
+                    "Base revision changeset could not be fetched from Phabricator, "
+                    "looking for existing issues based on the current date",
+                    task=settings.try_group_id,
+                )
+
+            # Mark know issues to avoid publishing them on this patch
+            self.find_previous_issues(issues, base_rev_changeset)
+            new_issues_count = sum(1 for issue in issues if issue.new_issue is True)
```
La0, line 117: Nit

```suggestion
            new_issues_count = sum(issue.new_issue for issue in issues)
```

bot/code_review_bot/workflow.py
```diff
@@ -288,6 +322,41 @@ def index(self, revision, **kwargs):
...
+        assert (
+            self.backend_api.enabled
+        ), "Backend storage is disabled, comparing issues is not possible"
+
+        current_date = datetime.now().strftime("%Y-%m-%d")
+
+        # Group issues by path, so we only list know issues for the affected files
+        groups = defaultdict(list)
+        for i in issues:
+            groups[i.path].append(i)
```
La0, line 341: Just use itertools.groupby

La0 (changes requested):

bot/code_review_bot/workflow.py
vrigal: I'm not sure paths are consecutive here (e.g. it would create multiple groups if different analyzers detected issues on the same file). I noticed groupby is used somewhere else to group by analyzer.

vrigal:
I ran the rebased branch with my local setup, results seems coherent (though I did not pulled the correct changeset for my local mozilla-central clone -that is referenced on try-, meaning hashes are computed incorrectly). 

Example message 1 :
```
Code analysis found 201 defects in diff [706131](https://phabricator.services.mozilla.com/differential/diff/706131/):
 - 3 defects found by test-manifest (Mozlint)
 - 1 build error and 4 defects found by clang-tidy
 - 10 defects found by eslint (Mozlint)
 - 183 defects found by clang-format (Mozlint)

WARNING: Found 194 defects (warning level) that can be dismissed.

IMPORTANT: Found 7 defects (error level) that must be fixed before landing.
[…]
---
7 errors and 855 warnings were detected but already existing on the mozilla-central repository.
```

Example message 2:
```
Code analysis found 203 defects in diff [706445](https://phabricator.services.mozilla.com/differential/diff/706445/):
 - 3 defects found by eslint (Mozlint)
 - 3 defects found by test-manifest (Mozlint)
 - 1 build error and 4 defects found by clang-tidy
 - 192 defects found by clang-format (Mozlint)

WARNING: Found 202 defects (warning level) that can be dismissed.

IMPORTANT: Found 1 defect (error level) that must be fixed before landing.
[…]
---
2 errors and 854 warnings were detected but already existing on the mozilla-central repository.
```

bot/code_review_bot/workflow.py
La0: Can you sort by path here first, then group by ?

bot/code_review_bot/report/phabricator.py
```diff
@@ -304,95 +319,56 @@ def build_comment(
...
-        defects_details = ""
-        # Only display the hint when the revision has parents in his stack
-        rev_parents_count = len(self.api.load_parents(revision.phabricator_phid))
-        external_failures_count = rev_parents_count and sum(
-            1
-            for issue in issues
-            if issue.is_publishable() and not revision.contains(issue)
-        )
-        if nb == 1 and external_failures_count == 1:
-            defects_details = " (in a parent revision)"
```
marco-c, line 369: Is there a way we could detect if the issue was introduced by a parent revision instead of removing support for it?

bot/code_review_bot/revisions.py
marco-c: Just adding a comment here to explain why this was marked as resolved: by calling `random.seed` again, it's like resetting the randomness. So the series of calls `random.seed(X)` + `random.random` always generates the same result.
Please correct me if wrong.

marco-c: BTW, might be worth to add an explanatory comment in the code

bot/code_review_bot/workflow.py
```diff
@@ -95,8 +92,36 @@ def run(self, revision):
...
+        # Analyze issues in case the before/after feature is enabled
+        if revision.before_after_feature:
+            logger.info("Running the before/after feature")
+            # Search a base revision from the decision task
+            decision = self.queue_service.task(settings.try_group_id)
+            base_rev_changeset = (
+                decision.get("payload", {}).get("env", {}).get("GECKO_BASE_REV")
+            )
+            if not base_rev_changeset:
+                logger.info(
```
marco-c, line 105: Might be worth sending to sentry so we know how frequent this is?

bot/code_review_bot/workflow.py
```diff
@@ -197,6 +219,11 @@ def publish(self, revision, issues, task_failures, notices, reviewers):
         # Only publish errors and "in patch" warnings due to a backend timeout
         publishable_issues = [i for i in issues if i.is_publishable()]
 
+        if revision.before_after_feature:
+            # In case before/after feature is enabled, only store issues that are new
+            # That is required to specify the number of unresolved or closed issues
+            publishable_issues = [i for i in publishable_issues if i.new_issue]
```
marco-c, line 225: Currently we show in-patch issues even if they are not new, maybe we should keep that behavior (so developers could fix them as ride-along to their changes).

bot/code_review_bot/report/phabricator.py
```diff
@@ -191,9 +198,11 @@ def publish(self, issues, revision, task_failures, notices, reviewers):
             if patch.analyzer.name not in self.analyzers_skipped
         ]
 
-        if issues:
+        # Publish a statistic about issues already known on mozilla-central
+        known_issues = [issue for issue in issues if issue.new_issue is False]
+        if publishable_issues:
             # Publish detected patch's issues on Harbormaster, all at once, as lint issues
-            self.publish_harbormaster(revision, issues)
+            self.publish_harbormaster(revision, publishable_issues)
```
marco-c, line 205: Doesn't this mean publishing all issues, even previously existing ones?

bot/code_review_bot/report/phabricator.py
```diff
@@ -443,11 +430,31 @@ def pluralize(word, nb):
...
+        if revision.before_after_feature and known_stats:
+            known_issues_msg = " and ".join(
+                self.pluralize(level, nb) for level, nb in known_stats.items()
+            )
+            plural = sum(known_stats.values()) > 1
+            known_issues_msg = (
+                known_issues_msg
+                + f" {'were' if plural else 'was'} detected but already existing on the mozilla-central repository.\n"
+            )
+            comment += known_issues_msg
```
marco-c, line 450: Not sure it's worth adding this additional information as it isn't actionable if we don't give details.
Perhaps we could add it in the future once we have a way in the frontend to show the full list of pre-existing issues on a given revision.
Wdyt @abpostelnicu?

bot/code_review_bot/report/phabricator.py
vrigal: Maybe I'm wrong, but I assumed before/after actually do that, by handling failures outside the diff.
The current code ensures that the Phab revision has parents (I don't know what it exactly means) and count every issue outside the patch as issue from a parent revision.

bot/code_review_bot/workflow.py
vrigal: I never saw any task definition missing those values, I'm not sure this is even possible.
I removed that line so we are sure it raises the error.

bot/code_review_bot/report/phabricator.py
vrigal: This is related to https://github.com/mozilla/code-review/pull/1581#discussion_r1183178491

I moved the publication logic (depending on before/after) to `Issue.is_publishable()` so it is much cleaner now : issues inside the patch OR unknown to the backend.

bot/code_review_bot/workflow.py
vrigal: I discussed with Bastien, warnings are automatically raised as sentry errors (from tools module).

vrigal:
I could run the publication example from https://phabricator.services.mozilla.com/D175125?id=703755#5776158 with the latest commit (28d32cc678831bce20138927cd02476e4d7d3bf1), with diffs mentioned [here](https://github.com/mozilla/code-review/pull/1581#issue-1608907899), here are the logs:

## Without before/after:
### Diff 1
```
Code analysis found 2 defects in diff [703755](https://phabricator.services.mozilla.com/differential/diff/703755/):
 - 2 defects found by eslint (Mozlint)

WARNING: Found 1 defect (warning level) that can be dismissed.

IMPORTANT: Found 1 defect (error level) that must be fixed before landing.

You can run this analysis locally with:
 - `./mach lint --warnings --outgoing`

---
If you see a problem…
```

## Diff 2
```
Code analysis found 1 defect in diff [704194](https://phabricator.services.mozilla.com/differential/diff/704194/):
 - 1 defect found by eslint (Mozlint)
1 defect unresolved and 1 defect closed compared to the previous diff [703755](https://phabricator.services.mozilla.com/differential/diff/703755/).

WARNING: Found 1 defect (warning level) that can be dismissed.

You can run this analysis locally with:
 - `./mach lint --warnings --outgoing`

---
If you see a problem…
```

## With before/after:
### Diff 1
```
Code analysis found 68 defects in diff [704194](https://phabricator.services.mozilla.com/differential/diff/704194/):
 - 68 defects found by eslint (Mozlint)

WARNING: Found 68 defects (warning level) that can be dismissed.

You can run this analysis locally with:
 - `./mach lint --warnings --outgoing`

---
419 warnings were detected but already existing on the mozilla-central repository.
If you see a problem…
```

### Diff 2
```
Code analysis found 15 defects in diff [704194](https://phabricator.services.mozilla.com/differential/diff/704194/):
 - 15 defects found by eslint (Mozlint)

WARNING: Found 15 defects (warning level) that can be dismissed.

You can run this analysis locally with:
 - `./mach lint --warnings --outgoing`

---
472 warnings were detected but already existing on the mozilla-central repository.
If you see a problem…
```

I think something is still wrong with before/after (the normal publication works as expected here):
* There should be 9 issues and not 68 in the first diff with before/after
* Missing resolved/closed count (maybe I did something wrong)

I should probably add more complex tests for before/after, but it is expensive.

vrigal:
Those issues are marked as "new" (and not "in patch"), so it seems rational.
I probably had a setup problem. I can find those issues in the backend but hashes do not match (check: 'mozilla/reject-scriptableunicodeconverter', path: 'toolkit/mozapps/update/tests/data/shared.js).

bot/code_review_bot/report/phabricator.py
marco-c: On Phabricator you can have revisions that are linked one another, you can see the whole thing as a PR with multiple commits.
The first commit of the PR is based on a changeset from the repository (in our case, mozilla-central), the other commits of the PR are based on the previous commits of the PR.

Say we have this situation:
- changeset X on mozilla-central
- revision A, based on changeset X
- revision B, based on revision A

When we analyze A, using before/after, we find issues that are new compared to changeset X, and we publish them.
When we analyze B, we will publish the same issues and with this code we were saying that some of them were actually due to A and not B (see https://github.com/mozilla/code-review/pull/1243 and linked issues/PRs). In an ideal world we'd use before/after also between B and A, but that might be too difficult to do now and so we could just keep the message for now.

vrigal: As discussed during today's meeting, handling parent revisions would be more complex (requires to find all the parents revs, then use the endpoint to list existing issues by path and compare them). 
it would be nice to do this in a followup MR though.

vrigal: I opened a new issue, feel free to edit or add details : https://github.com/mozilla/code-review/issues/1705

marco-c: OK, WFM.

PR #1591, opened 2023-03-07
Store base and head revision references

### LINKED ISSUE #1523: Remove Phabricator required IDs on Revision model
This is a followup on #1522 

The bot cannot process all revisions from Mozilla Central as some of them (backouts) do not have the Phabricator revision comment in their description.

Right now, the backend require to know the associated revision on Phabricator (both numercial ID and internal PHID) to publish any issue.

We could make these fields nullable on `Revision` and add these fields:
- Primary key becomes a random UUID (from uuid4)
- nullable bugzilla ID (provided from HGMO by the bot)
- nullable mercurial hash (would only be provided for MC, as we have the `Diff` system for try)

This would also require to support all these fields on the endpoint to create new revisions.

vrigal:
Closes #1523

### DIFF
--- a/backend/code_review_backend/issues/admin.py
+++ b/backend/code_review_backend/issues/admin.py
@@ -22,8 +22,8 @@ class DiffInline(admin.TabularInline):
 
 
 class RevisionAdmin(admin.ModelAdmin):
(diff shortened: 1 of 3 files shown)


EvaBardou:
Closed in favor of #1600

PR #1593, opened 2023-03-09
docupload: add extra files check (bug 1819368)

zzzeid:


zzzeid:
> Would it be possible to also have some sort of test?

Done.

bot/tests/test_docupload_task.py
```diff
@@ -39,6 +39,21 @@
...
+NOTE: 4 documentation files were modified in diff 42
+
+They can be previewed for one week:
+- file [docs/folderA/index.rst](http://firefox-test-docs.mozilla.org/section1/index.html)
+
+- file [docs/folderB/folderC/index.rst](http://firefox-test-docs.mozilla.org/section2/index.html)
+
+- file [docs/folderB/folderC/subfolder/index.md](http://firefox-test-docs.mozilla.org/section2/subfolder/index.html)
+
+- file [mots.yaml](http://firefox-test-docs.mozilla.org/index.html)
```
zzzeid, line 52: I noticed that the link path leads to the root page -- this is somewhat understandable since we're not detecting the actual change in the documentation file itself. This should be improved to map to the actual doc file, however I'd like to avoid hardcoding the URL (e.g., in [direct_doc_url](https://github.com/mozilla/code-review/pull/1593/files#diff-774da06547f30fc28ca6c29ce9a4c274e67c173fbad29f840dd69087644cec3dR33)) since the doc path is defined in `mots.yaml` itself.

zzzeid: [Done](https://bugzilla.mozilla.org/show_bug.cgi?id=1822123).

PR #1600, opened 2023-03-13
Store base and head revision references

EvaBardou:
Refs #1523


La0:
Thanks for taking over. As discussed here is a list of things to do/build:
- test migration on DB dump
   - I'm restoring our (outdated) prod dump on a test server
- identify all cases to support on the data migration (try, MC, autoland, nss, ...?)
- expose new mercurial infos in API serializer (read)
- also require these informations on the API endpoint to create a revision
- display both repos & revisions in the frontend on list of diffs
- update the bot `Revision` initialisation (both from try & decision task) to retrieve the `GECKO_...` environment variable, store them on the instance
- ... and use them when creating a new revision on the backend

bot/code_review_bot/revisions.py
```diff
@@ -206,17 +208,22 @@ def from_decision_task(task: dict, phabricator: PhabricatorAPI):
         """
         Build a revision from a Mozilla decision task (e.g. from Autoland or Mozilla-central)
         """
-        assert task["payload"]["env"]["GECKO_HEAD_REPOSITORY"] in (
+        # Load repositories
+        repository = task["payload"]["env"]["GECKO_HEAD_REPOSITORY"]
+        target_repository = task["payload"]["env"]["GECKO_BASE_REPOSITORY"]
+
+        assert repository in (
```
La0, line 215: You can actually remove this assertion now !

bot/code_review_bot/revisions.py
```diff
@@ -206,17 +208,22 @@ def from_decision_task(task: dict, phabricator: PhabricatorAPI):
...
             REPO_AUTOLAND,
             REPO_MOZILLA_CENTRAL,
         ), "Decision task must be on autoland or mozilla-central"
 
-        # Load mercurial revision
+        # Load mercurial revisions
         mercurial_revision = task["payload"]["env"]["GECKO_HEAD_REV"]
```
La0, line 221: Same thing: please use `head_changeset` and `base_changeset`  to avoid confusion between mercurial revision and `Revision` instances, but also between base & head. Target is too broad

La0 (changes requested):

bot/code_review_bot/revisions.py
La0: Actually you can't as it's only using GECKO vars

EvaBardou:
I ran this implementation to ingest, using the **bot**, a `try` task and `autoland`/`mozilla-central` decision tasks, here are the resulting objects in the **backend**

### `try` task
https://firefox-ci-tc.services.mozilla.com/tasks/by5wuUUSRPaUofMH2y_T3Q

Capture d’écran du 2023-03-20 14-37-09

### `autoland` decision task
https://firefox-ci-tc.services.mozilla.com/tasks/HaJ3sL67SaGwtfT4_031oA

Capture d’écran du 2023-03-20 17-07-38

### `mozilla-central` decision task
https://firefox-ci-tc.services.mozilla.com/tasks/egKGy8o8Tjqk_32NiNLjMw

Capture d’écran du 2023-03-20 17-11-09

marco-c:
The data looks reasonable to me, only one question: why is the base repository for try "mozilla-central" and not "mozilla-unified"?

EvaBardou:
> The data looks reasonable to me, only one question: why is the base repository for try "mozilla-central" and not "mozilla-unified"?

It's due to [this logic](https://github.com/mozilla/code-review/blob/master/bot/code_review_bot/revisions.py#L420), where we search for the `phabricator_repository` configuration in the settings, if I'm not mistaken, it's to support all try/base combinations easily but I'm not 100% familiar with this part of the code.

La0 (changes requested):
Code is a LOT better when using consistent naming, thank you for that.

I'm adding one last change regarding the `setup_try` as we discussed a bit previously and @marco-c mentionned it too.

bot/code_review_bot/revisions.py
```diff
@@ -422,22 +430,28 @@ def setup_try(self, task_group_id, tasks):
                 assert (
```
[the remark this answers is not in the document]
EvaBardou: `setup_try` is already called only for try tasks, never for decision ones. :thinking: 

Marco's comment was about an erroneous value on a try task, the decision ones are as expected.

La0 (approved):
I was able to ingest revisions locally :ok_hand:

PR #1614, opened 2023-03-27
Use the parent decision task to initialize revisions from try

### LINKED ISSUE #1601: Use repository slug as Primary key instead of phabricator ID
Followup of #1523 

This is the same approach to generalize the backend without requiring Phabricator IDs.

We already have a uniqe slug on the Repository model, both phabricator IDs should be made nullable. All relations must be updated before-hand in a data migration.

All the endpoints must also be updated, and finally the bot too (as it references repositories during revision creation).

### LINKED ISSUE #1612: Use decision task to initialize revisions from try
Followup of #1600 (specific [comment](https://github.com/mozilla/code-review/pull/1600#issuecomment-1481160605))

We can now remove `Revision.setup_try` and update the `Revision.from_try_task` by loading the task definition from `settings.try_group_id` (as the decision task ID is also the group ID).

The goal is then to populate all 4 head & base revisions + changesets from the environment variable.

The configuration must be updated to declare the `decision_env_suffix` (which will be `GECKO`, `COMM`, ...).

The bot then looks for the presence of all 4 variables matching a suffix, then use these variables. If no variables are found, the bot must crash explicitly.

The `repository_try_name` is simply the last part of the **head** repo after last `/` (for `https://hg.mozill.org/try`, this is `try`).

EvaBardou:
Closes #1612

### DIFF
--- a/bot/code_review_bot/revisions.py
+++ b/bot/code_review_bot/revisions.py
@@ -145,7 +145,7 @@ class Revision(object):
         return "Phabricator #{} - {}".format(self.diff_id, self.diff_phid)
 
     @staticmethod
-    def from_try_task(try_task: dict, phabricator: PhabricatorAPI):
+    def from_try_task(try_task: dict, decision_task: dict, phabricator: PhabricatorAPI):
         """
         Load identifiers from Phabricator, using the remote task description
         """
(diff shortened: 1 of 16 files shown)


La0 (changes requested):
Looks good, but as discussed you cannot remove the `repositories` configuration as it's required by events.

Please add a `decision_env_prefix` key on every item of the `repositories` list instead of using a single key for all prefixes

bot/code_review_bot/revisions.py
```diff
@@ -188,6 +188,52 @@ def from_try_task(try_task: dict, phabricator: PhabricatorAPI):
...
+                or head_changeset_key not in decision_env
+                or base_changeset_key not in decision_env
+            ):
+                continue
+
+            head_repository = decision_env[head_repository_key]
+            base_repository = decision_env[base_repository_key]
+            head_changeset = decision_env[head_changeset_key]
+            base_changeset = decision_env[base_changeset_key]
+            repository_try_name = head_repository.rstrip("/").rsplit("/", 1)[-1]
```
La0, line 217: We do not want to use the maxsplit=1 here (in the case of `hg.mozilla.org/ci/ci-configuration-try` we want `ci-configuration-try` not `ci/ci-configuration-try`). 

A better approach would be to use urrlib.parse.urlparse, then get the path attribute, then split on `/`, remove the empty strings, and keep the last one.

Code golf:

```
list(filter(None, urllib.parse.urlparse(head_repository).path.split('/')))[-1]
```

La0 (commented):
Just a nit on url parsing

Issue #1541, opened 2023-02-09
List known issues from mozilla-central and report new ones


La0:
This is a part of before/after.

The bot must use the endpoint from #1540 using the base revision of the current stack, also mentionning the creation date of the patch (as `date` parameter). We only want to use the `mozilla-central` repository for now.

Every file listed by the analyzers (that may contain files that are not in-patch) must be provided to this endpoint: all issues from the current analysis that are unknown to the backend should then be reported to the end user

:warning: we need to work out how we will communicate to the developer (message in current comment, an extra comment, mention in patch , ...?)

This feature should be set behing a feature flag, set from the configuration (as Taskcluster secret). the feature flag is a ratio (between 0.0 and 1.0), specify the a threshold to reach from a random function (0.3 means a patch has a 30% chance of getting this feature to run).

marco-c:
What if we show these issues just like we currently show the "in-patch" ones?

La0:
Small note after merging the backend endpoint to list known issues (#1540 ): we need to know the phabricator ID of the base revision

I suggest to implement that workflow into a method on the bot `Workflow` class, named `find_previous_issues`, which would take the list of current `Issue`  as input.

Full workflow for before/after in the bot would be (after `find_issues` is called in the bot main workflow):
1. check feature flag:
   - run `random.random()`
   - if it's above the feature flag value, stop the before/after analysis
2. search phabricator diffs for base revision (using mercurial hash from env variable `GECKO_BASE_REV` from the target tasks)
   - if a diff is found, use that as `revision` query parameter
   - otherwise just set the date to today
3. list all the files reported by the analyzers
   - :warning: this is a superset of the files from the patch, you'll need to iterate over all analyzer's issues and build a set of their path
4. for each path, query the new api with the date + optional revision ID
  - iterate over the current issues (passed as input parameter)
     - set boolean flag `new_issue` to True when the issue is not in the list from the backend (according to file + hash matching)

Now we have the information on which issue is new and we can report it in the phabricator comment:
- if an issue is `new_issue` it should always be reported, no matter its file path
- if the file path is outside of the patch, mark it explictly in the comment in a dedicated sections listing all issues outside of patch

La0:
Side note: this should only be enabled for try pushes !

PR #1624, opened 2023-03-28
Optional Phabricator references on the Revision model

### LINKED ISSUE #1523: Remove Phabricator required IDs on Revision model

vrigal:
Closes #1523

### DIFF
--- a/backend/code_review_backend/issues/migrations/0009_alter_revision_phid.py
+++ b/backend/code_review_backend/issues/migrations/0009_alter_revision_phid.py
@@ -15,7 +15,7 @@ def populate_uuids(apps, schema_editor):
     Postgres support generating UUIDs from the backend directly.
     """
     if "postgresql" in schema_editor.connection.vendor:
-        schema_editor.execute("INSERT INTO issues_revision id SELECT gen_random_uuid()")
+        schema_editor.execute("UPDATE issues_revision SET id = gen_random_uuid()")
     else:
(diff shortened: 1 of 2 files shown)


vrigal:
The migration (e241d06c73fd3835c48955fd5257a9c1a83feb3e) passed with SQLite (local DB) but failed with Postgres (production dump).
It needs some rework as currently Postgres complains about multiple PKs :
> django.db.utils.ProgrammingError: multiple primary keys for table "issues_revision" are not allowed
real    112m14.690s

The time needed to prepare the data is > 2h though (with a baremetal server), so it is a rather large migration.

Otherwise the frontend seems to work well and probably do not need any update. Revisions are displayed by their UUID, which is a bit unreadable.

vrigal:
We decided stop working on that migration to focus on the before/after feature (#1541) for now

PR #1660, opened 2023-04-17
Add prettier linter to pre-commit

marco-c:


marco-c:
CI will fail if we merge prettier without the changes

PR #1669, opened 2023-04-18
Draft implementation to use optional Phab references on the Revision model

EvaBardou:
Refs #1523

Based on Valentin's work from #1624


vrigal:
I have taken that work, and finished the implementation in !1670, sorry for the inconvenience of closing multiple issues.

PR #1693, opened 2023-05-03
bot: add specific request headers.

### LINKED ISSUE #1692: code-review bot should pass a User-Agent string for HTTP requests
Looking through the code-review bot codebase, I can see that the bot hits hg.mozilla.org with many different requests, including against the `raw-file` endpoint. We have experienced high load on the hg service due to many requests hitting the service at once, and it is difficult to distinguish legitimate traffic from crawlers and other bots. We know the code-review bot is hitting hgmo with many requests, but it is difficult to tell how many requests since the bot doesn't pass a User-Agent header. The code-review bot should be passing an identifiable User-Agent string everywhere possible.

I attempted to write a PR to pass the UA everywhere the code-review service makes an HTTP request, by moving the `TREEHERDER_HEADERS` constant into the common `code_review_tools` package and adding it to every request via the `headers=` kwarg. I haven't been able to make the tests run against this patch, but I can post it for reference in case that's useful.

abpostelnicu:
This fixes #1692 but also some imported libraries need to be updated like libmozdata.

### DIFF
--- a/bot/code_review_bot/config.py
+++ b/bot/code_review_bot/config.py
@@ -12,6 +12,7 @@ import os
 import shutil
 import tempfile
 
+import pkg_resources
 import structlog
 
 REPO_MOZILLA_CENTRAL = "https://hg.mozilla.org/mozilla-central"
@@ -26,6 +27,10 @@ RepositoryConf = collections.namedtuple(
     "RepositoryConf",
     "name, try_name, url, decision_env_prefix",
 )
+
+
+def GetAppUserAgent():
+    return {"user-agent": "code-review-bot/{}".format(settings.version)}
 
 
 class Settings(object):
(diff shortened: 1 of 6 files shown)


bot/code_review_bot/config.py
```diff
@@ -17,6 +17,8 @@
 REPO_MOZILLA_CENTRAL = "https://hg.mozilla.org/mozilla-central"
 REPO_AUTOLAND = "https://hg.mozilla.org/integration/autoland"
 
+HTTP_REQ_HEADERS = {"user-agent": "code-review-bot/1.0.0"}
```
sylvestre, line 20: maybe use code review version?

calixteman (approved):
LGTM.

PR #1715, opened 2023-05-11
When mentioning a patch has been rebased on top of m-c, mention the changeset hash

### LINKED ISSUE #447: Display the base revision used to apply patches
That would be pretty helpful for debugging issues when a patch fail to apply.
It should be displayed on our frontend, with a link toward that revision.

That means events should create the diff on the backend in a "working" state.

We could also add a new Build model to track phabricator build, as we can indeed get multiple execution per diff (not sure it's worth it here)

marco-c:
Fixes #447

Depends on https://github.com/mozilla/libmozevent/pull/100


marco-c:
Rebased!

PR #1741, opened 2023-05-25
bot: Add special case for generated files when dealing with their paths.

### LINKED ISSUE #1495: Frequent "Failed to download a file with an issue" warning on files from the build directory
The bot is trying to retrieve the file associated to an issue to build the hash, but this fails when the issue is in a generated file. E.g. GET https://hg.mozilla.org/try/raw-file/f233bd9d5d8f5b412505b48ef6e364d6c44a2304/obj-x86_64-pc-linux-gnu/layout/style/ServoStyleConsts.h [404].

abpostelnicu:
There are two cases that need special attention:
1.  ~~path is a symlink, so path must follow the symlink~~
2. path is to a generated file so it must be excluded from hash compilation and thus from publishing it on backend.

This fixes #1495

This patch addresses mainly to clang-tidy generated issues that can be traced back to generated files that are not actionable but also when we run `./mach static-analysis check ...` maybe we want to display these sort of issues.
That's why this has been fixed here and not in m-c.


marco-c:
> path is a symlink, so path must follow the symlink

Could this be fixed in m-c instead?

bot/code_review_bot/__init__.py
```diff
@@ -167,33 +167,39 @@ def hash(self):
...
+            local_repository = settings.runtime.get("mercurial_repository")
+            if local_repository:
+                logger.debug("Using the local repository to build issue's hash")
+                try:
+                    with open(local_repository / self.path) as f:
+                        file_content = f.read()
+                except (FileNotFoundError, IsADirectoryError):
                     logger.warning(
-                        "Failed to download a file with an issue", path=self.path
+                        "Failed to find issue's related file", path=self.path
```
marco-c, line 183: After logging this warning, we still need to set "file_content = None" as before.

bot/code_review_bot/__init__.py
```diff
@@ -167,33 +167,39 @@ def hash(self):
...
-            except ValueError:
-                # Build the hash with an empty content in case the path is erroneous
-                file_content = None
-            except requests.exceptions.HTTPError as e:
-                if e.response.status_code == 404:
+        # Build the hash only if the file is not autogenerated.
+        # An autogenerated file resides in the build directory that it has the
+        #  format `obj-x86_64-pc-linux-gnu`
+        file_content = None
+        if "/obj-x86_64-pc-linux-gnu/" not in self.path:
```
marco-c, line 174: Should we do:
```suggestion
        if "/obj-" not in self.path:
```
to catch more cases?

marco-c (approved):

PR #1804, opened 2023-07-04
Keep sentry-sdk to 1.22.2 for the moment

abpostelnicu:
sentry-sdk 1.23 and after has some issues with our code. COMMON_RECORD_ATTRS is not exposed directly in interations. 
The introduction of loguru to the sdk broke our integration.


marco-c (dismissed):
Sorry I totally missed this

PR #1881, opened 2023-09-05
backend: add new migration for revision.id field

abpostelnicu:


backend/code_review_backend/issues/migrations/0010_alter_revision_id.py
```diff
@@ -0,0 +1,22 @@
...
+        ("issues", "0009_revision_optional_phab_references"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="revision",
+            name="id",
+            field=models.BigAutoField(primary_key=True, serialize=False),
+        ),
+    ]
```
marco-c, line 22: Isn't this operation the same as this one (https://github.com/mozilla/code-review/blob/8ef38fa655fa55926eb62c6defdba845683b87de/backend/code_review_backend/issues/migrations/0009_revision_optional_phab_references.py#L47-L51) in the previous migration?

marco-c (approved):

PR #1890, opened 2023-09-12
backend: redo the django db clean up script.

abpostelnicu:
The db clean up script has some limitations by using delete on a filtered QuerySet, django stores all filtered elements in memory from different tables that are cross referenced by FKs.
If we are to run this script weekly this is not an issue since the number of issues and issuelinks will be limited to a couple of thousands or at most tens of thounsads, but in the initial cleanup this is suboptimal since we currently have around 40mil issues and 24mil links, thus leading to an OOM from the backend.
The proposal is to first cleanup the link with _raw_delete and after that the issues with _raw_delete.
Using _raw_delete we avoid storing the enteties in the memory thus avoiding the OOM and the delete is performed in a single query. This avoids the CASCADE protection so each time the Django model is updated, these filteres and the logic how they are called need to be in sync.

### DIFF
--- a/backend/code_review_backend/issues/tests/commands/test_cleanup_issues.py
+++ b/backend/code_review_backend/issues/tests/commands/test_cleanup_issues.py
@@ -105,7 +105,7 @@ class CleanupIssuesCommandTestCase(TestCase):
     def test_cleanup_issues_default_nb_days(self):
         self.assertEqual(Issue.objects.count(), 6)
         with self.assertLogs() as mock_log:
-            with self.assertNumQueries(12):
+            with self.assertNumQueries(6):
                 call_command(
                     "cleanup_issues",
                 )
@@ -125,14 +125,14 @@ class CleanupIssuesCommandTestCase(TestCase):
             [
                 f"{LOG_PREFIX}Retrieved 1 old revisions from either autoland or mozilla-central to be deleted.",
                 f"{LOG_PREFIX}Page 1/1.",
-                f"{LOG_PREFIX}Deleted 4 issues.IssueLink, 1 issues.Diff, 1 issues.Revision, 2 issues.Issue.",
+                f"{LOG_PREFIX}Deleted 4 IssueLink, 1 Diff, 2 Issue, 1 Revision.",
             ],
         )
 
     def test_cleanup_issues_custom_nb_days(self):
         self.assertEqual(Issue.objects.count(), 6)
         with self.assertLogs() as mock_log:
-            with self.assertNumQueries(12):
+            with self.assertNumQueries(6):
                 call_command("cleanup_issues", "--nb-days", "4")
 
         self.assertEqual(Issue.objects.count(), 2)
@@ -146,6 +146,6 @@ class CleanupIssuesCommandTestCase(TestCase):
             [
                 f"{LOG_PREFIX}Retrieved 2 old revisions from either autoland or mozilla-central to be deleted.",
                 f"{LOG_PREFIX}Page 1/1.",
-                f"{LOG_PREFIX}Deleted 6 issues.IssueLink, 1 issues.Diff, 2 issues.Revision, 4 issues.Issue.",
+                f"{LOG_PREFIX}Deleted 6 IssueLink, 1 Diff, 4 Issue, 2 Revision.",
             ],
         )


backend/code_review_backend/issues/tests/commands/test_cleanup_issues.py
```diff
@@ -123,27 +124,27 @@ def test_cleanup_issues_default_nb_days(self):
         self.assertListEqual(
             mock_log.output,
             [
+                f"{LOG_PREFIX}Didn't find any old issues to delete.",
```
marco-c, line 127: Shouldn't this test be deleting old issues?

backend/code_review_backend/issues/management/commands/cleanup_issues.py
```diff
@@ -29,24 +29,39 @@ def add_arguments(self, parser):
...
         rev_to_delete = Revision.objects.filter(
             base_repository__slug__in=["autoland", "mozilla-central"],
             head_repository__slug__in=["autoland", "mozilla-central"],
             created__lte=clean_until,
         )
 
-        count = rev_to_delete.count()
-        if not count:
+        count_rev = rev_to_delete.count()
+        if count_rev:
```
marco-c, line 59: Let's keep the old behavior of early return to minimize the diff and to keep the rest of the code with a smaller indentation

backend/code_review_backend/issues/management/commands/cleanup_issues.py
```diff
@@ -29,24 +29,38 @@ def add_arguments(self, parser):
...
-        count = rev_to_delete.count()
-        if not count:
+        if not count_rev:
             logger.info("Didn't find any old revision to delete.")
             return
 
-        # Delete revisions, with their related IssueLink and Diff
+        issuelinks_to_delete = IssueLink.objects.filter(revision_id__in=rev_to_delete)
+        issues_to_delete = Issue.objects.filter(
+            id__in=issuelinks_to_delete.values("issue_id")
```
vrigal, line 49:
```suggestion
            issue_links__revision_id__in=rev_to_delete
```
Saves a request and potentially avoid an OOM by storing all the IDs from the IssueLink table.

vrigal:
This approach looks fine to me ! I think however few improvements are possible. \
* Using `.delete()` instead of `._raw_delete()`. It should also perform a single request for Issue/IssueLink and avoids using an internal method.
* Deleting diffs before the revisions, as for the others FKs (so they also appear in the statistics).

EvaBardou:
I'm afraid that with this new implementation you'll sometimes delete `Issue` objects wrongfully.

Issues are linked to (one or multiple) revisions through `IssueLink` model. Previously, we deleted revisions (and `IssueLink`/`Diff` with the `CASCADE` clause) and searched for orphan issues, not linked to any revision anymore, to clean them up.

Now, you:
- fetch all `IssueLink` objects linked to the revisions to be deleted
- fetch all `Issue` objects linked to those links (:warning: even if they could be linked to other revisions too!)
- delete all fetched objects

You should probably:
- fetch `IssueLink` objects linked to the revisions to be deleted, delete them
- fetch `Diff` objects linked to the revisions to be deleted, delete them
- delete the revisions
- clean up potential orphan issues (this will avoid deleting issues which are still linked to other revisions)

vrigal:
Yes, but Eva is right we cannot simply delete issues as they can be referred in newer revisions.
`Diff` has a cascade relation to `Revisions`, so it can be something like this IMO:
* Drop IssueLink from old revisions
* Drop Diffs related to those revisions
* Drop the Revisions
* Drop Issues that are not linked anymore to a revision via the IssueLink M2M (so referenced issues are not deleted)

vrigal:
Discussion summary:
* There are only revisions with base repo = `mozilla-central` and head repo = `try` on the SQL dump.
* Postgres run OOM during the deletion of IssueLink entries
    * `mem=1279M(250.0%)`
    * Probably because of the joins with `revision` and `repository` tables ? (The OOM happens with an empty qs)

Maybe it is possible to use chunks in order to perform the delete in two steps to avoid a join:
* Fetching revisions IDs for this chunk
* Filtering issues (and diff) from their FK to one of these IDs

```python=
DEL_CHUNK_SIZE = 10000

rev_to_delete = Revision.objects.filter(
    base_repository__slug="mozilla-central"
    head_repository__slug="try",
    created__lte=clean_until,
)
revs_count = rev_to_delete.count()
for start in range(0, revs_count, DEL_CHUNK_SIZE):
    # First fetch revisions IDs in a first DB request
    chunk_rev_ids  = rev_to_delete.order_by('id')[start : start + UPDATE_CHUNK_SIZE].values_list('id', flat=True)
    # Delete IssueLink for this chunk
    links_count, _ = IssueLink.objects.filter(revision_id__in=chunk_revs_ids).delete()
    # Delete Diff for this chunk
    diffs_count, _ = Diff.objects.filter(revision__id__in=chunk_revs_ids).delete()
# Finally, delete issues that are not linked to any revision anymore
issues_count, _ = Issue.objects.filter(issue_links=None).delete()
```

backend/code_review_backend/issues/management/commands/cleanup_issues.py
```diff
@@ -29,24 +32,49 @@ def add_arguments(self, parser):
...
-        _, stats = rev_to_delete.delete()
-        # Delete issues that are not linked to a revision anymore
-        _, issues_stats = Issue.objects.filter(revisions__isnull=True).delete()
-        stats.update(issues_stats)
+
+        stats = {}
+        for start in range(0, count_rev, DEL_CHUNK_SIZE):
+            # First fetch revisions IDs in a first DB request
+            chunk_rev_ids = rev_to_delete.order_by("id")[
+                start : start + UPDATE_CHUNK_SIZE
```
vrigal, line 58:
```suggestion
                start : start + DEL_CHUNK_SIZE
```
We should keep a single variable here, otherwise it could cause some IDs to be skipped.

backend/code_review_backend/issues/management/commands/cleanup_issues.py
```diff
@@ -29,24 +32,51 @@ def add_arguments(self, parser):
...
-        _, stats = rev_to_delete.delete()
-        # Delete issues that are not linked to a revision anymore
-        _, issues_stats = Issue.objects.filter(revisions__isnull=True).delete()
-        stats.update(issues_stats)
+
+        stats = {}
+
+        iterations = math.ceil(count_rev / DEL_CHUNK_SIZE)
+        i = 0
+        for start in range(0, count_rev, DEL_CHUNK_SIZE):
```
vrigal, line 55: @abpostelnicu you have a very useful [enumerate](https://docs.python.org/3/library/functions.html#enumerate) function to do that in Python:
```suggestion
        for i, start in enumerate(range(0, count_rev, DEL_CHUNK_SIZE), start=1):
```
`start=1` allows to start from index 1 at the first iteration (instead of zero)

backend/code_review_backend/issues/management/commands/cleanup_issues.py
```diff
@@ -29,24 +32,54 @@ def add_arguments(self, parser):
...
-        # Delete issues that are not linked to a revision anymore
-        _, issues_stats = Issue.objects.filter(revisions__isnull=True).delete()
-        stats.update(issues_stats)
+
+        stats = {}
+
+        iterations = math.ceil(count_rev / DEL_CHUNK_SIZE)
+        i = 0
+        for i, start in enumerate(range(0, count_rev, DEL_CHUNK_SIZE), start=1):
+            i += 1
```
vrigal, line 56:
```suggestion
```
No need to iter on `i`, this is done automatically by the `enumerate` function

vrigal:
I was able to run the current version (db17bbf8a555fcada3e3c43653e222e9687f31aa) against the testing DB with a fresh dump :
```
Retrieved 3198 old revisions from either autoland or mozilla-central to be deleted.
Page 1/7.
Page 2/7.
Page 3/7.
Page 4/7.
Page 5/7.
Page 6/7.
Page 7/7.
Deleted 231322 IssueLink, 3357 Diff, 231322 Issue, 3198 Revision.
```
It took 25 minutes to run, and use a bit of RAM (~1GB) on the machine I ran the script.

vrigal:
Please not that it does not remove all Issue entries that are not linked via the IssueLink M2M.
Migration `issue 0007` was not applied completely and a lot of issues are unlinked.
They could probably be cleaned up manually.

vrigal:
14 million, the DB says :
```
SELECT COUNT(*) AS "__count"
  FROM "issues_issue"
  LEFT OUTER JOIN "issues_issuelink"
    ON ("issues_issue"."id" = "issues_issuelink"."issue_id")
 WHERE "issues_issuelink"."id" IS NULL

Execution time: 232.740169s [Database: default]
Out[1]: 14140445
```
Migration `issues 0007` links issues via the IssueLink M2M from the most recent to the oldest. So I suppose these are the oldest of all the issues.

vrigal (approved):

PR #1891, opened 2023-09-17
bot: add the option to specify checker that shouldn't use before/after

abpostelnicu:
Add the ability to disable before/after for specific checkers based on config secrets.

### DIFF
--- a/bot/code_review_bot/__init__.py
+++ b/bot/code_review_bot/__init__.py
@@ -134,7 +134,7 @@ class Issue(abc.ABC):
         This allow publishing issues based on other criteria, like in_patch.
         """
         if taskcluster.secrets.get(
-            f"{self.display_name.upper()}_DISABLE_PUBLICATION_BEFORE_AFTER", False
+            f"{self.analyzer.name.upper()}_DISABLE_PUBLICATION_BEFORE_AFTER", False
         ):
             return False
 


bot/code_review_bot/__init__.py
```diff
@@ -127,6 +127,19 @@ def build_extra_identifiers(self):
         """
         return {}
 
+    @property
+    def allow_before_and_after(self):
+        """
+        Allow the possibility for an issue to avoid being published based on before/after.
+        This allow publishing issues based on other criteria, like in_patch.
+        """
+        if taskcluster.secrets.get(
```
vrigal, line 136: I think this condition should rather be implemented in `code_review_bot.revisions.Revision.before_after_feature`, so it also disable the feature in `code_review_bot.workflow`. Otherwise some issues will not be published but still analyzed with the feature and so fail comparing with previous issues.

vrigal:
It would be nice to add a test for this flag, maybe in `tests.test_workflow`

bot/code_review_bot/__init__.py
vrigal: After looking closely at the publication part, this should work as expected until those settings changes between the base revision and the analyzed revision (issues reported as new, or closed with before/after because not published in the backend).
The workflow is:
Issue detected -> marked as new if not published before -> not published -> not mentioned on Phabricator
Ideally it would be:
Issue detected -> marked as new if not published before -> published -> not mentioned on Phabricator

Otherwise I suggest to simply rename `<ANALYZER>_DISABLE_BEFORE_AFTER` to `<ANALYZER>_DISABLE_PUBLICATION` to make it clear. The check can be done in the `is_publishable` method directly also.

bot/code_review_bot/__init__.py
```diff
@@ -127,6 +127,19 @@ def build_extra_identifiers(self):
...
         return {}
 
+    @property
+    def allow_before_and_after_publish(self):
+        """
+        Allow the possibility for an issue to avoid being published based on before/after.
+        This allow publishing issues based on other criteria, like in_patch.
+        """
+        if taskcluster.secrets.get(
+            f"{self.display_name.upper()}_DISABLE_PUBLICATION_BEFORE_AFTER", False
```
vrigal, line 137:
```suggestion
            f"{self.analyzer.name.upper()}_DISABLE_PUBLICATION_BEFORE_AFTER", False
```
In the code `issue.name` == `issue.analyzer.display_name` == `issue.analyzer.name`, but we should rather pick the original name of the analyzer here (from the task definition).

vrigal:
A little suggestion, otherwise this looks good to me.
I was able to test based on phab rev [D188806](https://phabricator.services.mozilla.com/D188806) (with `TRY_TASK_GROUP_ID=DWrMLVdXTxm9CHCg7HqR9Q` and `TRY_TASK_ID=IXK6HhIcSXS29XZWJhsMgA`).

Original Phabricator comment:
```
Code analysis found 15 defects in diff [766323](https://phabricator.services.mozilla.com/differential/diff/766323/):
 - 2 defects found by stylelint (Mozlint)
 - 13 defects found by eslint (Mozlint)

WARNING: Found 13 defects (warning level) that can be dismissed.

IMPORTANT: Found 2 defects (error level) that must be fixed before landing.
```

I faked the fact these warnings are not in-patch (so they are not published) with `SOURCE-TEST-MOZLINT-ESLINT_DISABLE_PUBLICATION_BEFORE_AFTER: yes` in my secrets:
```
Code analysis found 2 defects in diff [766323](https://phabricator.services.mozilla.com/differential/diff/766323/):
 - 2 defects found by stylelint (Mozlint)

IMPORTANT: Found 2 defects (error level) that must be fixed before landing.
```

vrigal (approved):

PR #1895, opened 2023-09-20
Use podman rather img to build images

petemoore:
The `img` tool seems to be no longer actively maintained. Fuzzing also switched to podman in https://github.com/MozillaSecurity/orion/pull/242.

Credit to @matt-boris for the idea, finding the `BUILD_TOOL` env var, and discovering the fuzzing PR!

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -47,7 +47,7 @@ tasks:
         then: "https://api.code-review.testing.moz.tools"
         else: "https://api.code-review.moz.tools"
 
-    taskboot_image: "mozilla/taskboot:0.3.6"
+    taskboot_image: "mozilla/taskboot:0.4.1"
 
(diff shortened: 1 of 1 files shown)


marco-c:
Maybe we should actually make `podman` the default in task-boot?

petemoore:
> Maybe we should actually make `podman` the default in task-boot?

Agreed! I did this here to check it works before switching, and needing a new task-boot release.

matt-boris:
Just created https://github.com/mozilla/task-boot/pull/390. I think what we can change in this repo would be setting https://github.com/mozilla/code-review/blob/e0da094d42f85e6c168e6bd953d4caa31919e018/.taskcluster.yml#L50 to the `latest` tag, _or_ if it should still be pinned to an exact version, upgrade after the new version of taskboot is released.

matt-boris (approved):
Nice 👍🏻

marco-c (approved):
Thanks!

PR #1901, opened 2023-09-28
Update taskcluster-hook.json

matt-boris:
Add generic-worker cache scope

Fixes failures like this https://community-tc.services.mozilla.com/tasks/BtnJ7KFrR6qKl9umM2MjEA/runs/0.


marco-c (approved):
Thanks!

PR #1920, opened 2023-10-10
Use checks api

lotas:
Switch to checks API. 

Also test build pipeline, why tasks are deadline expired


lotas:
> @lotas what is this suppose to do?

Github initially had Statuses API only, that was the default option, where jobs would be attached to the commit directly (iirc) and were quite limited.
They later introduced Checks API, and it has more features. For example those runs can now be seen on the PR level: 
https://github.com/mozilla/code-review/pull/1920/checks (extra tab) and also allows to attach extra information like description (task logs) and annotations that you could see in the github UI.
It would also allow to use `re-run` from github UI to retry failed tasks

lotas:
We fixed community-tc config here: https://github.com/taskcluster/community-tc-config/pull/688/files 
so this project can use extra scope `queue:route:checks`
And FxCI tasks seems to be running just fine, so should be safe yes. 

Merged similar on code-coverage also https://github.com/mozilla/code-coverage/pull/2112

PR #1923, opened 2023-10-12
test checks and pr

lotas:
to see if pr can be merged since statuses api is not used anymore


lotas:
@abpostelnicu I just wanted to test statuses. 

So after I added checks-api reporting in .taskcluster.yml, there were few changes - taskcluster now uses Checks to report build statuses, instead of Statuses.

This leads to this situation:
image

Which means those rules were "hardcoded" in branch protection rules. Someone with admin access can verify this. 
They are no longer required, as those statuses no longer run, but checks run instead. 

Those branch protection rules [support both statuses and checks](https://docs.github.com/en/enterprise-cloud@latest/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks) so just removing hardcoded names should probably help.

Sorry for the extra work 🙈

marco-c:
I've fixed the branch protection rules.

lotas:
thanks @marco-c !

PR #2041, opened 2024-01-16
events: Ensure all configured repositories exist on Phabricator

marco-c:


events/code_review_events/workflow.py
```diff
@@ -119,17 +119,28 @@ def _build_conf(config):
...
             return config
 
-        repositories = {
-            phab_repo["phid"]: Repository(_build_conf(conf), cache_root)
-            for phab_repo in self.api.list_repositories()
-            for conf in repositories
-            if phab_repo["fields"]["name"] == conf["name"]
-        }
-        assert len(repositories) > 0, "No repositories configured"
+        repository_mapping = {}
```
suhaibmujahid, line 122: nit: I would keep the dict comprehension and then check the size:
```py
if len(repository) != len(repository_mapping):
    # find the missed repos if that is important info to log
    missed_repos = set(r["name"] for r in repositories) - set(r.name for r in repository_mapping.values())
    # raise an exception
```

suhaibmujahid (dismissed):

Issue #2193, opened 2024-05-29
Add Taskcluster cache for local Mercurial clones


marco-c:
In order to actually use the work done in https://github.com/mozilla/code-review/pull/1517, we need to enable local cloning.
In order to enable local cloning, we need to make sure we don't clone from scratch in each task, but we rely on a Taskcluster cache.

La0:
I'm looking into this and have a few questions:
- there already exists [firefox source code checkouts](https://firefox-source-docs.mozilla.org/taskcluster/caches.html#version-control-caches) in Taskcluster cache, but they are restricted to tasks ran by run-task: would you consider moving the code-review bot execution to run-task ?
- if using custom caches for the code review bot, we would need one cache per repository. 
   - when the cache is empty, some code-review bot tasks would need to populate it (initial clone), and will take longer: is that ok ?

La0:
I noticed that we already have [cache enabled](https://github.com/mozilla/code-review/blob/master/integration/taskcluster-hook.json#L30) for the Integration hook.

Unfortunately the [hook on Community TC](https://community-tc.services.mozilla.com/tasks/X-st65oCSg61OmEzwxGpIg/runs/0/logs/public/logs/live.log) does not successfully run anymore due to a missing secret.

We could restore that functionality, and test the clone & re-use from cache functionalities from that hook before updating runtime tasks

WDYT @marco-c ?

La0:
I spent a bit of time today on this topic and noticed a few things:
- the integration test is broken because of invalid Taskcluster proxy url  and automated conversion from docker-worker to generic-worker. As I do not know what's the long term plan regarding this hook & its viability, I stopped looking into it.
- I made a first PR to control directly the mercurial cache behaviour from the CLI args, removing Taskcluster "production" check, and simplifying the overall code. In the end, the taskcluster cache is usable because I removed the tempdir path creation. As `robust_checkout` is already used, the cache will be automatically purged & setup at the correct version on each run
- The mercurial version tools extension is not at the latest version (but just lack 1 whitespace changeset)
- The mercurial version is outdated (6.4.5 in code  vs 6.7.4 on pypi)
- I have issues running locally related to sentry-sdk (probably an invalid version locally that has changed some modules)

Next steps:
- get a review from @Archaeopteryx or @marco-c 
- enable cache on https://hg.mozilla.org/ci/ci-configuration/file/tip/hooks/project-relman/code-review-testing.yml

marco-c:
[...]

PR #2235, opened 2024-07-01
Simpler controls for mercurial local clone

La0:
Refs #2193 

This PR brings simpler controls for mercurial clone so it becomes usable on production.

- removed configuration `runtime` update
- extend existing `settings.setup()` to configure a mercurial cache destination
- do not create temporary path for mercurial clone so we can rely on reusable taskcluster cache
- use directly `robust_checkout` method


bot/code_review_bot/config.py
```diff
@@ -124,21 +134,29 @@ def on_production(self):
         """
         return self.app_channel == "production" and self.taskcluster.local is False
 
-    def is_allowed_path(self, path):
```
marco-c, line 127: git diff is pretty bad here, I wonder why!

marco-c (dismissed):

marco-c (approved):

PR #2289, opened 2024-07-30
restrict on push builds to `master` and `production` in taskcluster

bhearsum:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1907217 we're making scopes more explicit for branches. Because of this, pushes to branches not named in that repository will not be able to run. If this will be problematic, or we need to add more branches to this list, please let me know and we'll sort it out!

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -66,7 +66,7 @@ tasks:
       then: linux-gw-gcp
       else: generic-worker-ubuntu-24-04
   in:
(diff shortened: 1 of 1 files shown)


marco-c:
@bhearsum we also have a "testing" branch that we'd need to add to the list.

bhearsum:
@marco-c - we had discussed on Slack using L1 and L3 pools depending on the branch. I ended up not doing that, because all of the tasks fired by hooks run on the same workers anyways, so there's no benefit. We should use leveled workers for those tasks at some point, but it's probably very low priority.

marco-c (approved):
Thanks!

PR #2343, opened 2024-09-09
Add recent setuptools as dependency to support setup with Python 3.12 after upgrade to Django 5.1

Archaeopteryx:


La0:
Please write the same version number everywhere @vrigal

vrigal (approved):

marco-c (approved):

PR #2372, opened 2024-09-16
Add backend docker build github action

La0:


La0:
@marco-c @Archaeopteryx I'm trying to add a simple github action here (to build the Backend docker image for the GCP migration). But the Github Workflow does not start.

Could you look into the Github settings if you can somehow enable the github actions (probably disabled due to our use of Taskcluster CI).

The goal is NOT to replace Taskcluster CI, simply to build the docker images in Github actions so we can use Cloudops tools to publish on GCP docker repo

Archaeopteryx:
That's a question for Marco, GH doesn't show me the Settings tab.

marco-c:
> @marco-c @Archaeopteryx I'm trying to add a simple github action here (to build the Backend docker image for the GCP migration). But the Github Workflow does not start.
> 
> Could you look into the Github settings if you can somehow enable the github actions (probably disabled due to our use of Taskcluster CI).

It looks like they are already enabled:
immagine

> The goal is NOT to replace Taskcluster CI, simply to build the docker images in Github actions so we can use Cloudops tools to publish on GCP docker repo

The tools are only available in the GH actions context?

La0:
> The tools are only available in the GH actions context?

My understanding is that if we build the docker image through github actions, we'll get more support & tool integration from cloudops

marco-c:
> > The tools are only available in the GH actions context?
> 
> My understanding is that if we build the docker image through github actions, we'll get more support & tool integration from cloudops

Let's figure out what it means practically. Since we have to keep Taskcluster as CI anyway for ease of integration with Firefox CI, avoiding the need for two CI systems would be ideal.

La0:
Closed in favor of #2377

Issue #2392, opened 2024-10-02
Upgrade Vue.js 2 to 3


La0:
Refs #2259 

This is a large work, fortunately the frontend code base is relatively small.

Steps, in successive Pull requests, following [recommendations](https://v3-migration.vuejs.org/recommendations.html):
1. Start  [Vue 3 migration build](https://v3-migration.vuejs.org/migration-build.html) to update Vue.js dependencies
   - reach step 10 where both vuex & vue-router are upgraded
2. Migrate from vuex (now deprecated) to Pinia
   - follow [migration guide](https://pinia.vuejs.org/cookbook/migration-vuex.html)
   - you can probably keep everything in a single store as there are few actions
3. [upgrade vue-router](https://router.vuejs.org/guide/migration/index.html) to support vue 3
4. Switch to vite instead of webpack for build tool as it's the [recommended default now](https://vuejs.org/guide/scaling-up/tooling.html)
5. Iterate over individual warnings from vue js migration build until the compat mode can be disabled

mlbonhomme:
Now that I don't have errors anymore I think I need access to the Heroku group or something https://github.com/mlbonhomme/code-review/blob/master/docs/debugging.md#heroku so I can get data in the backen to actually test the frontend?

La0:
@mlbonhomme I'll give you access to our internal DB for this purpose

La0:
We'll just do the vue chart, while trying to migrate data storage onto Treeherder

PR #2397, opened 2024-10-02
Migrate frontend to vue3 (part 1)

mlbonhomme:
refs #2392 

based on #2414 

- compat build
- no more `this.$set` or filters
- I wanted to take a look at what [the stats page](https://code-review.moz.tools/#/stats) is supposed to look like but it doesn't seem to work ? so hopefully it looks like what it's supposed to look like


La0 (changes requested):
There remains 2 `this.$set` in stats component.

Otherwise this works relatively well against a dockerized backend & recent data from `load_issues`.

Could you split the chart-js migration into another PR (to merge first)  ? Or it's a lot of work ?

PR #2413, opened 2024-10-07
simplify dependabot.yml

Archaeopteryx:
Feel free to reject if the labels are needed for the pull requests.


marco-c:
We haven't been following any rule, we should decide if we want to have labels or prefixes in the PRs

marco-c (approved):

Issue #2436, opened 2024-10-11
Remove single issue creation


La0:
Refs [Code review issues in Treeherder](https://docs.google.com/document/d/199Jd9Hoc17h-Y8M4-UuR-L7FEh0Ct-rBOuGQJi4PVqE/edit)

The bulk creation endpoint should be the only one able to create issues.
This is a quick win to simplify bot `publish_issues` method to only support the existing bulk endpoint (when `bulk > 0` path remains)

Then remove the endpoint exposed as `/diff/{diff_id}/issues/` so that only the revision bulk issues creation remain.

La0:
This data migration should be a [lot more performant](https://github.com/mozilla/code-review/blob/e16a48e41fa9d3ca7ed0bf2ba18d6f36916f68c7/backend/code_review_backend/issues/migrations/0012_move_issues_attributes.py)

It renders as 

```
UPDATE "issues_issuelink" 
SET 
"in_patch" = (
  SELECT U0."in_patch"
  FROM "issues_issue" U0
  WHERE U0."id" = ("issues_issuelink"."issue_id")
  ORDER BY U0."created" ASC LIMIT 1
),
"new_for_revision" = (
  SELECT U0."new_for_revision" 
  FROM "issues_issue" U0 
  WHERE U0."id" = ("issues_issuelink"."issue_id")
  ORDER BY U0."created" ASC LIMIT 1
)
```

PR #2447, opened 2024-10-14
De-duplicate issues

### LINKED ISSUE #2449: De-duplicate issues by hash
Depends #2448 

We want to de-duplicate issues by hash, as the attributes `in_patch` & `new_for_revision` have been moved to `IssueLink`.

This will remove a lot of duplicated issues (on my local tests down from 163k to 3k).

Steps:
- data migration removing duplicated issues by hash and updating their links
   -   updating the links also needs a de-duplication as there is a unique constraint on revision+diff+issue
- Update the bulk issues creation endpoint to respect that constraint, skipping existing issues (just ignoring conflicts should be ok, while creating new links).

La0:
Closes #2449 

This brings unique `Issue.hash` to the backend by adding a de-duplicating data migration and a new constraint.

The migration iterates over all issues with duplicated hashes, and groups all the linked `IssueLink` to a single one `Issue` so that the constraint can be applied.
The migration is restartable, and will skip all previously de-duplicated `Issue`: this is required to process large database such as prod


vrigal:
> https://github.com/mozilla/code-review/issues/2436#issuecomment-2411620098

I think the above comment is related to this issue. I will continue the work on #2451 without the deduplication.

Archaeopteryx:
@La0 Could you rebase on top of the now landed #2451 ?

Archaeopteryx (approved):

PR #2450, opened 2024-10-14
Add mozdata.ini for libmozdata to bot and events to set user-agent

Archaeopteryx:
Required for requests with libmozdata since version 0.2.8

### DIFF
--- a/bot/mozdata.ini
+++ b/bot/mozdata.ini
@@ -1,0 +1,2 @@
+[User-Agent]
+name = code-review-bot/1.8.4
--- a/events/mozdata.ini
+++ b/events/mozdata.ini
@@ -1,0 +1,2 @@
+[User-Agent]
+name = code-review-events/1.8.4
--- a/bot/docker/Dockerfile
+++ b/bot/docker/Dockerfile
@@ -6,6 +6,7 @@ ADD tools /src/tools
 RUN cd /src/tools && python setup.py install
 
 ADD bot /src/bot
+COPY /src/bot/mozdata.ini mozdata.ini
 RUN cd /src/bot && python setup.py install
 
 # Add mercurial & robustcheckout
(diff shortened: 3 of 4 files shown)


bot/mozdata.ini
```diff
@@ -0,0 +1,2 @@
+[User-Agent]
+name=mozilla/code-review/bot
```
marco-c, line 2: It'd be ideal to follow the same User-Agent we're setting at https://github.com/mozilla/code-review/blob/978802d2dd451462bb4b62977f7d3fe326312d5d/bot/code_review_bot/config.py#L32

events/mozdata.ini
```diff
@@ -0,0 +1,2 @@
+[User-Agent]
+name=mozilla/code-review/events
```
marco-c, line 2: Here too, the same style

marco-c:
Is this enough to have the config available?

marco-c:
I see you have the commit on testing, but on top of 1.8.3. You should test it on testing on top of 1.8.4.
I have a hunch it won't be enough to have the file in the top directories

La0:
This patch works when using the bot in a python virtualenv, but not in docker.

As libmozdata looks for the `mozdata.ini` in current working directory, you need to:
- `COPY bot/mozdata.ini /mozdata.ini`
- or symlink it from the already copied `/src/bot/mozdata.ini` there
- or set `WORKDIR /src/bot`

We could also use libmozdata `ConfigEnv` to set that value using environment variable built into the config or Dockerfile

bot/mozdata.ini
```diff
@@ -0,0 +1,2 @@
+[User-Agent]
+name = code-review-bot/1.8.4
```
marco-c, line 2: We'll have to remember to update this when we release new versions. Maybe we could add a placeholder here and replace it with a command in Dockerfile?

WDYT @La0 ?

La0: It would actually be simpler to implement our own `Config` class that simply provides the user-agent value, as that's the only one we want to provide.

I'll make a small PR to show this, then you can chosse

La0: I made #2452 to use a Local configuration

marco-c:
Closing in favour of #2452.

PR #2451, opened 2024-10-15
 Move Issue non hashed attributes to IssueLInk

### LINKED ISSUE #2448: Move Issue in_patch & new_for_revision attributes to IssueLInk
Depends #2445 

To be able to deduplicate issues (by adding a `UNIQUE` on `Issue.hash`), we need to move the `Issue.in_patch` & `Issue.new_for_revision`

You can steal this [commit](https://github.com/mozilla/code-review/pull/2447/commits/6a673ebf06f9b156aad7f71257f0c4f34fcf51fb) to get the data migration (please add tqdm instead of the percent displayed, I did that in a [followup](https://github.com/mozilla/code-review/blob/fadde98d9d3862cd67e0f6839c2b94edb3c100a8/backend/code_review_backend/issues/migrations/0012_move_issues_attributes.py)). I needed it to start working on unique hashes.

The migration is just the first part, you need to then update:
- issue serializer, so that you can still pass the values, but they are stored on IssueLink instead of Issue directly
- `load_in_patch` django command
- DiffViewSet queryset
- IssueCheckDetails queryset
- IssueCheckStats queryset
- IssueViewSet perform_create
- a bunch of unit tests...

The goal is to avoid changing the Rest API interface, so that the frontend & bot do not need any changes.

vrigal:
Closes #2448

This PR moves the following attributes from `Issue` model up to `IssueLink`:
- `new_for_revision`
- `in_patch`
- `line`
- `nb_lines`
- `char`

These attributes are **not in the hash payload**, so they should not be on the `Issue` model as we want to set a unique constraint on the hash later on (see #2447 )

The API endpoints remain compatible with the previous implementation so that the bot can simply provide a list of attributes that will result in creation of Issue+IssueLink

Both `load_issues` command & the frontend remain working as exepcted


vrigal:
Almost done, still few failures detected from tests:
> FAILED (failures=1, errors=3)

La0:
I'll check if we need to move other attributes in this PR (line positions).

La0 (commented):
I'll simply say that I'm not the best reviewer here as I contributed a good chunk of the code with @vrigal

Archaeopteryx (approved):

PR #2475, opened 2024-10-28
Add duplicated issues tests for bulk entities creation endpoint

### LINKED ISSUE #2474: Bulk issues creation does not lookup existing hashes
The bulk issues creation endpoint does not currently lookup for pre-existing issues with same hashes.

We can see some duplicate issues trigger a [500 on testing](https://mozilla.sentry.io/issues/6015636066/?)

I thought this was covered in the recent changes, but apparently we do not have a test case re-publishing issues with the same hash.

We need to:
1. add a unit test that hit the [issues publication endpoint](https://github.com/mozilla/code-review/blob/master/backend/code_review_backend/issues/api.py#L227) twice:
   - once with issue A & B (should have no problem)
   - second time with issue A & C (it should crash on master)
2. replace `Issue.objects.bulk_create` in [IssueBulkSerializer](https://github.com/mozilla/code-review/blob/master/backend/code_review_backend/issues/serializers.py#L265) by a loop using `Issue.objects.get_or_create`
   - we did the same thing in [load_issues](https://github.com/mozilla/code-review/blob/master/backend/code_review_backend/issues/management/commands/load_issues.py#L74)

mlbonhomme:
closes #2474 

But I only added the tests; I do not have access to Sentry so I can't see what caused the 500 errors or when, but duplicated issues do seem to be handled by the current code with 

```
issues = Issue.objects.bulk_create(
            [Issue(**values) for values in validated_data["issues"]],
            ignore_conflicts=True,
        )
```

but maybe I'm missing something.

### DIFF
--- a/backend/code_review_backend/issues/tests/test_api.py
+++ b/backend/code_review_backend/issues/tests/test_api.py
@@ -334,3 +334,127 @@ class CreationAPITestCase(APITestCase):
         link = issue.issue_links.get()
         self.assertFalse(link.new_for_revision)
         self.assertEqual(link.line, 1)
+
+    def test_create_issue_bulk_alread_exists(self):
+        """
+        You cannot try to create the same issue twice through the bulk endpoint
+        """
+        payload_1 = {
+            "issues": [
+                {
+                    "hash": "somemd5hash",
+                    "line": 1,
+                    "analyzer": "remote-flake8",
+                    "level": "error",
+                    "path": "path/to/file.py",
+                    "in_patch": True,
+                    "new_for_revision": False,
+                },
+                {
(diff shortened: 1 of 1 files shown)


La0:
This unit test does not cover the 500 case (rarer) but the 400 with duplicate issue hash.

It's happening a lot on testing so we need to fix that quickly. Here is a [sample task](https://firefox-ci-tc.services.mozilla.com/tasks/W4GXvHaRSZ-8C1b7IatqsA/runs/0/logs/public/logs/live.log) that failed to submit any issue on the backend

Archaeopteryx (approved):

PR #2502, opened 2024-11-04
Allow issues bulk creation with existing hashes

### LINKED ISSUE #2474: Bulk issues creation does not lookup existing hashes

vrigal:
Closes #2474

### DIFF
--- a/backend/code_review_backend/issues/tests/test_api.py
+++ b/backend/code_review_backend/issues/tests/test_api.py
@@ -213,7 +213,7 @@ class CreationAPITestCase(APITestCase):
         # Once authenticated, creation will work
         self.assertEqual(Issue.objects.count(), 0)
         self.client.force_authenticate(user=self.user)
-        with self.assertNumQueries(8):
+        with self.assertNumQueries(6):
             response = self.client.post(
                 f"/v1/revision/{self.revision.id}/issues/", data, format="json"
             )
@@ -294,7 +294,7 @@ class CreationAPITestCase(APITestCase):
             ],
         }
         self.client.force_authenticate(user=self.user)
-        with self.assertNumQueries(8):
+        with self.assertNumQueries(7):
             response = self.client.post(
                 f"/v1/revision/{self.revision.id}/issues/", data, format="json"
             )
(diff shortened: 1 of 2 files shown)


backend/code_review_backend/issues/tests/test_api.py
```diff
@@ -386,15 +395,18 @@ def test_create_issue_bulk_alread_exists(self):
...
-                {"issues": [{"hash": ["issue with this hash already exists."]}]},
-            )
+            self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertCountEqual(
+            [issue["hash"] for issue in response.json()["issues"]],
+            ["somemd5hash", "athirdmd5hash"],
+        )
+        new_issues = list(Issue.objects.order_by("created"))
+        self.assertEqual(len(new_issues), 3)
+        self.assertListEqual([i.id for i in issues], [i.id for i in new_issues[:2]])
```
La0, line 409: It would be cleaner to check the hash value, hardcoding the hashes there: we make 100% sure we only have expected non-duplicated values

La0 (changes requested):

La0 (approved):
Well done, this now works with repeated calls to the bot :+1:

Archaeopteryx (approved):

PR #2511, opened 2024-11-12
Also migrate char, line and nb_lines attributes to IssueLink

vrigal:
Follow up #2473


La0:
I was thinking that we could probably do a single SQL subquery by writing it directly in sql (should be faster instead of making 5 subqueries).

Something like

```sql
update issues_issuelink as link
set 
  in_patch=i.in_patch,
  new_for_revision=i.new_for_revision,
  char=i.char,
  line=i.line,
  nb_lines=i.nb_lines
from issues_issue as i
where i.id = link.issue_id

```

La0:
Closing in favor of #2513

PR #2512, opened 2024-11-12
Implement issues de-duplication in raw SQL

La0:
Even after the cleanup, the migration 0014 (issue deduplication) was still slow on testing DB.

First I noticed that we lacked an index on `Issue.hash` so I added that. Query runtime sped up significantly, but not enough (still requiring a week of processing).

Then I thought of creating temporary tables to write only the needed rows, but while doing that I realised I could simply partition the issues grouped by hashes, thus identifying duplicates, then update in-place the IssueLink references towards Issues, and finally delete the duplicate issues.

I'm currently running this on testing DB and on a copy of the production DB. I'll report back with times


La0:
This ran on a copy of production DB in 4h19min, allowing the unique constraint on `Issue.hash` to be applied

La0:
This also ran succesfully on Heroku testing DB

PR #2530, opened 2024-11-18
Automatically create repositories referenced on a revision

### LINKED ISSUE #2521: Create automatically repositories
We can simplify onboarding of new repositories publishing issues by automatically creating `Repository` instance from the `RevisionSerializer` (on both `base_repository` and `head_repository`).

To do that, we'll first need to simplify the model, removing the `PhabricatorModel` inheritance:
- remove `Repository.phid` field which is now unused (it was used at first to identify repositories upon creation but we rely onits url now)
- set `Repository.id` as an incremental integer

Then we can simply create new repositories from their provided urls:
- parse the url with `urllib.urlparse`
- checking they are hosted on `hg.mozilla.org` (as netloc from urlparse output)
- use path from urlparse output as `Repository.slug`
- let the DB provide a new increment for the ID

This should not change the bot code. Side-effect: `load_issues` command should work out of the box on an empty DB

vrigal:
Closes #2521


backend/code_review_backend/issues/serializers.py
```diff
@@ -26,20 +28,48 @@ class RepositorySerializer(serializers.ModelSerializer):
...
+            try:
+                return self.get_queryset().get(url=url)
+            except Repository.DoesNotExist:
+                self.fail("invalid_url")
+            except (TypeError, ValueError):
+                self.fail("invalid")
+                return
+        try:
+            repo, _ = self.get_queryset().get_or_create(
+                url=url, defaults={"slug": parsed.path}
```
La0, line 59: Please remove the leading `/` from `parsed.path` so we get `mozilla-central` instead of `/mozilla-central`

La0 (changes requested):
Could you also create the repository automatically from `load_issues` command when it's not in DB ?
Same rules as from the API
- must belong to the list of known repos
- slug from parsed path (stripping leading `/`)

La0 (commented):
Latest changes are OK, Repositories are created without leading `/` both from the API & load_issues.

Please implement the cleanup mentionned by @marco-c and we should be ready to merge

backend/code_review_backend/issues/migrations/0015_remove_repository_phid_alter_repository_id.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ("issues", "0014_unique_hash"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="repository",
+            name="phid",
+        ),
+        migrations.AlterField(
```
Archaeopteryx, line 16: Does the current value of the counter need to be set? From the Treeherder MySQL to PostgreSQL migration I remember the primary keys (sometimes?) having low values (starting at the initial value?) and storing new rows failed because of this. Is Django smart enough to calculate the current value?

backend/code_review_backend/issues/management/commands/load_issues.py
```diff
@@ -179,22 +181,35 @@ def load_local_reports(self):
...
 
+    def get_or_create_repository(self, url):
+        """Retrieve a repository or create it if its URL must match allowed hosts"""
+        parsed = urlparse(url)
+        if parsed.netloc not in settings.ALLOWED_REPOSITORY_HOSTS:
+            try:
+                return Repository.objects.get(url=url)
+            except Repository.DoesNotExist:
+                logger.warning(
+                    f"No repository exist with URL {url} "
```
Archaeopteryx, line 192: Should be `exists`

backend/code_review_backend/issues/management/commands/cleanup_issues.py
```diff
@@ -28,7 +34,20 @@ def add_arguments(self, parser):
...
+            head_revisions__isnull=True,
+            diffs__isnull=True,
+        )
+        return unused_repositories._raw_delete(unused_repositories.db)
+
     def handle(self, *args, **options):
+        stats = defaultdict(int)
+        repo_count = self.cleanup_repositories()
+        if repo_count:
+            stats["Repository"] = repo_count
```
La0, line 49: You can directly log all the repositories deleted, otherwise the log at end may not be reached if no issues is found

La0 (dismissed):

Archaeopteryx (approved):

backend/code_review_backend/issues/migrations/0015_remove_repository_phid_alter_repository_id.py
La0: Good catch, you're right, I tested on our local copy of prod data, and got an error because the sequence has not been initialized:

```python
>>> from code_review_backend.issues.models import *
>>> from django.db.models import *
>>> Repository.objects.all().aggregate(max=Max('id'))
{'max': 1001}
>>> Repository.objects.create(url='http://test', slug="test")
Traceback (most recent call last):
  File "/home/bastien/Envs/cr-backend/lib/python3.10/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "issues_repository_pkey"
DETAIL:  Key (id)=(1) already exists.
```

@vrigal you can add for postgresql the following sql query in the migration
```
SELECT setval(pg_get_serial_sequence('issues_repository', 'id'), coalesce(max(id)+1, 0), false) FROM issues_repository;
 setval 
```

La0 (approved):

Archaeopteryx (approved):

PR #2531, opened 2024-11-19
Release 1.9.0

Archaeopteryx:


frontend/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "code-review-frontend",
-  "version": "1.8.6",
+  "version": "1.8.7",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
```
La0, line 6: You need to set the version below too (in `""` first section)

La0 (commented):
I suggest to set the version to 1.9.0 as there are a lot of breaking changes (and no way to go back data-wise)

La0 (approved):

PR #2534, opened 2024-11-22
New smart API to list issues

### LINKED ISSUE #2533: New API endpoint to list issues
This is the last development in the backend before porting code over to Treeherder.

Creation endpoints (revision, diff, issues) are OK now, but the bot also needs a way to retrieve issues in 3 stages:
 - unresolved issues in a revision
 - known issues (issues from that diff that are already present in DB - linked to another issue)
 - closed issues (issues from previous diffs in the parent revision that are not linked to the targeted diff)

We can build a single endpoint that takes as URL parameters:
- Diff ID (it can be the PK as the bot requires these informations after creating its diff)
- mode `unresolved|known|closed`

The endpoint serializer returns each Issue ID & hash (that's it).

The Endpoint would be exposed as `/diff/{diff_id}/issues/{mode}/`

We'll then implement their usage in the bot in a followup

vrigal:
Based on #2530

Closes #2533

This will requires more tests to correctly handle all edge cases.

Actually all those detection are based on Issue's hashes. I noticed that mean the same issue (linter, level, message) can only be reported once per file by the bot.


vrigal:
I noticed the same combination of (analyzer, level, check, message) can only be reported once per file by the bot, which may lead to inconsistent returned values. For example, a new issues could be reported as known (if a similar issue is known on this file), or an issue returned as both unresolved and closed (multiple issues, only one has been closed).

I think we should make this endpoint KISS, and handle such edge cases in the bot.

backend/code_review_backend/issues/serializers_v2.py
```diff
@@ -0,0 +1,15 @@
+from rest_framework import serializers
+
+from code_review_backend.issues.models import IssueLink
+
+
+class IssueLinkHashSerializer(serializers.ModelSerializer):
+    """Serialize an Issue hash from the IssueLink M2M"""
+
+    id = serializers.UUIDField(source="issue_id")
```
vrigal, line 9: Note: add `read_only` on each field as it will not be applied from the Meta class.

backend/code_review_backend/issues/api_v2.py
```diff
@@ -0,0 +1,121 @@
...
+    @property
+    def known_issues(self):
+        """
+        Issues that have also being detected from mozilla-central.
+        """
+        return self.distinct_issues(
+            self.diff.issue_links.annotate(
+                known=Exists(
+                    IssueLink.objects.filter(
+                        revision__base_repository__slug="mozilla-central",
```
La0, line 56: You can use the curent diff base repository id to make the query generic & faster (one less join)

```suggestion
                        revision__base_repository_id=self.diff.revision.base_repository_id,
```

backend/code_review_backend/issues/api_v2.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def get_queryset(self):
+        return getattr(self, f"{self.mode}_issues").filter(**{self.mode: True})
+
+    def distinct_issues(self, qs):
+        """
+        Convert a list of issue links to unique couples of (issue_id, issue_hash)
+        """
+        attributes = ("issue_id", "issue__hash")
+        if "postgresql" in settings.DATABASES["default"]["ENGINE"]:
```
La0, line 43: Just assume you are running on Postgres, we'll remove support for sqlite

backend/code_review_backend/issues/api_v2.py
```diff
@@ -0,0 +1,121 @@
...
+        parent revision, and are still present on the current diff.
+        Filters out issues that are known by the backend.
+        """
+        previous_diff = (
+            self.diff.revision.diffs.filter(created__lt=self.diff.created)
+            .order_by("created")
+            .last()
+        )
+        if not previous_diff:
+            return IssueLink.objects.none().annotate(unresolved=Value("True"))
```
La0, line 76: It would be cleaner to raise a 400 with a clear exception message `No previous diff available.`

backend/code_review_backend/issues/api_v2.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+
+    @property
+    def closed_issues(self):
+        """
+        Issues that were listed on the previous diff of the same
+        parent revision, but does not exist on the current diff.
+        """
+        # Retrieve the previous diff
+        previous_diff = (
```
La0, line 95: You should create a property on the class for previous_diff

backend/code_review_backend/issues/api_v2.py
```diff
@@ -0,0 +1,121 @@
...
+                        diff=self.diff,
+                        issue=OuterRef("issue"),
+                    )
+                )
+            )
+        )
+
+
+urls = [
+    # Prevails on the generic issues endpoint
```
vrigal, line 115: Nit: remove this comment.

backend/code_review_backend/issues/api_v2.py
```diff
@@ -0,0 +1,121 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
La0, line 1: Please create a dedicated module for v2 api as `code_review_backend.issues.v2` where you'll have:
- `api.py` with the new endpoint + urls declaration
- `serializers`
- `tests` folder

La0 (changes requested):
A few comments to address as discussed together.

Then we'll run it on a local prod dump, and can probably ship to testing as it adds new endpoints.

La0:
You will also need to serialize the `previous_diff_id` when applicable (on unresolved & closed issues).

This does not invalidate the fact that a 400 should be raised when it's not present in these cases

backend/code_review_backend/issues/api_v2.py
vrigal: `revision__base_repository_id=self.diff.revision.base_repository_id` do includes all issues from all diffs with the base repository. I suppose it should match `revision__head_repository_id` instead ?

Issue #2124, opened 2024-03-21
don't create try pushes for old phabricator revisions


jcristau:
If reviewbot fails to create a try push for a given phabricator revision in some time (1 hour?), it should give up instead of holding on to it forever.

We're seeing an issue right now where reviewbot is delayed making try pushes by 2 days.

marco-c:
This would be fixed around here: https://github.com/mozilla/libmozevent/blob/a41fad391146bbdc182b3bcc295353c30f76bfbc/libmozevent/mercurial.py#L357.

marco-c:
We should return a failure and post to Phab that we run out of time.

La0:
- we should check the age of the [build there](https://github.com/mozilla/libmozevent/blob/a41fad391146bbdc182b3bcc295353c30f76bfbc/libmozevent/mercurial.py#L443) as there are other stop conditions
- the age is obtainable through `PhabricatorBuild.diff.[some path in JSON]`

La0:
Side note: we need this patch to test deployment of #2449 (if we need to get some downtime, it will spam with try pushes otherwise)

marco-c:
This is fixed in libmozevent, we just need to upgrade the lib.

La0:
The usage of this patch in events gives [this Sentry issue](https://github.com/mozilla/code-review/pull/2548)

La0:
I made a cleaner version of build expiry check, adding a new state on the build, and custom error messages.

These 2 PRs still require some unit tests

La0:
@Archaeopteryx could you review https://github.com/mozilla/libmozevent/pull/109 then we need to build a libmozevent release (should just be a tag push) and use it in #2550

marco-c:
Fixed by #2576.

PR #2550, opened 2024-11-26
Support build expiry in events

La0:
Refs #2124 

Requires https://github.com/mozilla/libmozevent/pull/109

This patch send a custom error message to Phabricator & Lando when a build has expired


La0:
Not needed anymore

Issue #2518, opened 2024-11-14
Don't wait for Phabricator events in the 'events' service, wait through a Taskcluster hook instead


marco-c:
Taskcluster now supports [triggering a webhook](https://docs.taskcluster.net/docs/reference/core/hooks/firing-hooks#webhooks) through a unique url (with a token).
We could build a new hook that is triggered by each Phabricator diff (same config as currently), creating a task associated to the diff. 
That task would run the same workflow as events:
- Clone base repo, using a Taskcluster cache and robustcheckout
- Apply the diff on tip
- Push to try
- Update phabricator harbormaster state for the diff

This would resolve the issues around disk space that we have on Heroku, allowing us to do things like #2004 and #2204.
This would also allow us to make the 'events' service slimmer. If we do both this and #945, we could get rid of the 'events' service altogether.

La0:
The overall workflow would be :

1. Phabricator triggers a webhook on Taskcluster
   - using a new hook declared in fxci-config repo
   - using Taskcluster recent [public webhook trigger](https://docs.taskcluster.net/docs/reference/core/hooks/firing-hooks#webhooks)
2. The hook triggers a new task with the request from the caller (in `payload`
```
{
    firedBy: "triggerHookWithToken",
    taskId: 'IgfFQSAqQwysozyeB7udBw',
    payload: {..}               // API call payload
}
```
We know [from our libmozevent implementation](https://github.com/mozilla/libmozevent/blob/master/libmozevent/phabricator.py#L24-L27) that Phabricator is able to send
- **diff ID** (that's the only required info)
- **target PHID** (required to send messages back to Phabricator
- repo PHID
- revision ID


3. From there we can reuse code from libmozevent, without using the Bus (used on events to run continously). Internal wokflow is the one mentionned by @marco-c above:
  - clone base repo using TC cache & robustcheckout  ([implemented in libmozevent](https://github.com/mozilla/libmozevent/blob/master/libmozevent/mercurial.py#L94))
  - apply the patch (also in [libmozevent](https://github.com/mozilla/libmozevent/blob/master/libmozevent/mercurial.py#L140) , just requires to create a [PhabricatorBuild](https://github.com/mozilla/libmozevent/blob/master/libmozevent/phabricator.py#L18) to match current implem)
  - push to try is [immediate](https://github.com/mozilla/libmozevent/blob/master/libmozevent/mercurial.py#L289)
[...]

PR #2576, opened 2024-12-11
Port events workflow to bot

La0:
Refs #2518 

This is the new code-review bot behaviour to support Phabricator webhook triggers, replacing the events system running on Heroku.

The revision & diffs are loaded using these environment variables:
-  `PHABRICATOR_OBJECT_PHID` which is the revision phid
- `PHABRICATOR_TRANSACTIONS` which is a json encoded list of Phabricator transactions, which contains the history of the revision at the time of the trigger

The bot must:
- retrieve the revision, repository, diff and Harbormaster target from Phabricator API
- clone the base repo at its tip
- load the stack of patches from Phabricator
- apply them on the clone
- push to try
- report state (or errors) on Phabricator
- report state (or errors) on Lando

The bot now depends on libmozevent, allowing the code port to be relatively minimal (no mercurial code, no Phabricator patch stack management).
But this means the bot needs to implement the high-level objects used by code-review-events to communicate with libmozevent (ie `PhabricatorBuild` to propagate state of a diff) and instantiate the various parts of events system (phabricator actions, mercurial repo & worker).

The CLI boot code is extended to support the new environment variable and run this new workflow (named `Analysis` here). In that mode, a `Revision` is instantiated using a dedicated method that:
1. load the revision from Phabricator
2. identify the repository against known configurations (to get try urls)
3. find the diff from Phabricator using provided transactions
4. find the Harbormaster build target (passing through buildable & build)

Once the `Revision` object is created, the workflow starts, running a new `start_analysis` method which will:
1. check the revision author is not blacklisted
2. check the mercurial cache is enabled
3. check the SSH key is available to push on try (from secrets)
4. Update phabricator state to Working
5. create the internal `RevisionBuild` which allows us to use libmozevent classes
6. instantiate libmozevent services:
   - `PhabricatorActions` to update the build state & fetch stack of patches
   - `Repository` & `MercurialWorker` to clone the repo & apply stack of patches, then push to try
7. Check the diff is public
8. Load stack of patches from Phabricator
9. Clone the repository
10. Apply the stack of patches & push to try when applying was successful
11. Publish outcome on Phabricator & Lando

We could work on followups to port subsystems directly into the bot, even removing the need for libmozevent (extras could be in libmozdata, especially for phabricator calls).

### DIFF
--- a/bot/code_review_bot/analysis.py
+++ b/bot/code_review_bot/analysis.py
@@ -113,7 +113,7 @@ def publish_analysis_phabricator(payload, phabricator_api):
                 namespace="code-review",
                 name="mercurial",
                 result=UnitResultState.Unsound,
-                details=f"WARNING: The base revision of your patch is not available in the current repository.\nYour patch has been rebased on central (revision {build.actual_base_revision}): issues may be positioned on the wrong lines.",
+                details=f"WARNING: The base revision of your patch is not available in the current repository.\nYour patch has been rebased on central (revision {build.actual_base_revision}): issues may be positioned at the wrong lines.",
             )
             phabricator_api.update_build_target(
                 build.target_phid, BuildState.Work, unit=[warning]
--- a/bot/code_review_bot/analysis.py
+++ b/bot/code_review_bot/analysis.py
@@ -26,22 +26,38 @@ class RevisionBuild(PhabricatorBuild):
         # Incremented on an unexpected failure during build's push to try
         self.retries = 0
 
+        # Revision used by Phabricator updates
+        # Direct output of Phabricator API (not the object passed here)
+        self.revision_url = None
+        self.revision = None
+
+        # Needed to update Phabricator Harbormaster
+        self.target_phid = revision.build_target_phid
+
         # Needed to load stack of patches
         self.diff = revision.diff
         self.diff_id = revision.diff_id
         self.stack = None
 
+        # Needed to apply patch and communicate on Phabricator
+        self.base_revision = None
+        self.actual_base_revision = None
+        self.missing_base_revision = False
+
     def __str__(self):
         return f"Phabricator Diff {self.diff_id}"
 
 
-def publish_results(self, payload):
-    if not self.publish:
-        logger.debug("Skipping Phabricator publication")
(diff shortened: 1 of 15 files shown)


bot/code_review_bot/workflow.py
```diff
@@ -215,6 +215,18 @@ def _build_tasks(tasks):
...
 
+    def start_analysis(self, revision):
+        """
+        Apply a patch on a local clone and push to try to trigger a new Code review analysis
+        """
+        logger.info("Patch should be applied here...")
+
+        # TODO: clone upstream at tip
+
+        # TODO: apply stack of patches
```
Archaeopteryx, line 226: The current implementation doesn't revert the repository properly if a patch fails to apply. Example: A patch is being applied and adds file2 and modifies file1. There is a conflict for file1. If the patch is applied again by another review bot run, it will fail because file2 already exists. This had been reported last week by Julien. Let me know if I shall file a issue (the logs of this example are inaccessible now).

La0: This should probably be a followup, I'm focusing now on porting the current workflow from events+libmozevent into the bot, by reusing as much existing structure as possible.

This PR will rely on libmozevent dependency to avoid re-implementing the whole clone+apply stack + push workflow

bot/code_review_bot/analysis.py
```diff
@@ -0,0 +1,185 @@
...
+    elif mode == "success":
+        if build.missing_base_revision:
+            # Publish a warning message on Phabricator in the Unit Tests section,
+            # as done for other warnings/errors from the bot, since this section
+            # is centered and at the top of the page.
+            warning = UnitResult(
+                namespace="code-review",
+                name="mercurial",
+                result=UnitResultState.Unsound,
+                details=f"WARNING: The base revision of your patch is not available in the current repository.\nYour patch has been rebased on central (revision {build.actual_base_revision}): issues may be positioned on the wrong lines.",
```
Archaeopteryx, line 116: `on the wrong lines` => `at the wrong lines`

Archaeopteryx (approved):

Archaeopteryx (approved):

PR #2605, opened 2025-01-24
Use most recent diff when not found in transaction

La0:
I published a [new test revision](https://phabricator-dev.allizom.org/D3704) on phabricator-dev, and the task triggered by Phabricator did not get any information about a diff in the transactions.

The [resulting task ](https://firefox-ci-tc.services.mozilla.com/tasks/Qckop4W6Sju4tM1UdQXlNQ) failed early with the message:

```
Failed to load revision        error='No DIFF found in transactions' phabricator=https://phabricator-dev.allizom.org/api/ task=None
```

This PR adds a new behaviour, to fallback on the most recent diff available on the revision.

### DIFF
--- a/bot/code_review_bot/revisions.py
+++ b/bot/code_review_bot/revisions.py
@@ -353,19 +353,31 @@ class Revision:
                 diff_phid = new
                 break
 
-        # Check a diff is found
+        # Check a diff is found in transactions or use last diff available
         if diff_phid is None:
-            raise Exception("No DIFF found in transactions")
+            diffs = phabricator.search_diffs(
+                revision_phid=revision_phid,
+                attachments={"commits": True},
+                order="newest",
+            )
+            if not diffs:
+                raise Exception(f"No diff found on revision {revision_phid}")
+            diff = diffs[0]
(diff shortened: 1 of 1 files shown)


Archaeopteryx (approved):

bot/code_review_bot/revisions.py
```diff
@@ -353,19 +353,31 @@ def from_phabricator_trigger(
                 diff_phid = new
                 break
 
-        # Check a diff is found
+        # Check a diff is found in transactions or use last diff available
         if diff_phid is None:
```
marco-c, line 357: How can this happen? Is there a risk we trigger a task, in the meantime the developer uploads a new patch, we trigger another one, and so in the end we have two tasks running for the same diff?

La0: I do not know why Phabricator did not send the Diff creation transaction in the list of transactions, but it was really missing.

You are right, there is currently no mutex-like mechanism to prevent running multiple times. We could probably achieve that by using Taskcluster routes (setting a route in a task, and checking it does not exist at task start).

PR #2611, opened 2025-02-06
Depend on Phabricator build target ID in trigger mode

La0:
This PR is needed to support the new payload that will be sent by Phabricatior's Harbormaster build step using https://github.com/mozilla-conduit/phabricator/pull/49

This simplifies a lot the `Revision` setup as we will directly use a Phabricator build target directly linked to a Diff & Revision. So there is no need to "guess" the diff from transactions and fallback to the last diff.

### DIFF
--- a/bot/code_review_bot/cli.py
+++ b/bot/code_review_bot/cli.py
@@ -163,10 +163,9 @@ def main():
             revision = Revision.from_decision_task(
                 queue_service.task(settings.mozilla_central_group_id), phabricator_api
             )
-        elif settings.phabricator_revision_phid:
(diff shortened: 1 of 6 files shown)


La0:
The testing hook has been updated to expose the environment variable `PHABRICATOR_BUILD_TARGET` when triggered with a hook token and a valid JSON payload.

@Archaeopteryx Could you review this PR ?
FYI here is a [manually triggered](https://firefox-ci-tc.services.mozilla.com/tasks/KgAxkcwjTYiFQmcolhNxmw/definition) task on the hook.

Archaeopteryx (approved):

PR #2652, opened 2025-03-05
Remove events source code

La0:
And mentions from docs.


La0:
As the events production system has been stopped today, we can land this .

marco-c (approved):

PR #2850, opened 2025-06-27
Skip already merged patches when building push to try

### LINKED ISSUE #2827: Code review bot tries to apply closed revisions as well as updated revisions
In this [revision](https://phabricator.services.mozilla.com/D252859) we saw in diff [1059155](https://phabricator.services.mozilla.com/differential/diff/1059155/) and [diff 1059166](https://phabricator.services.mozilla.com/differential/diff/1059166/) that code review bot was failing to apply the patch.

This was because it used to depend on [D252322](https://phabricator.services.mozilla.com/D252322), which had already been landed at the time the patch was later updated.

This causes confusion, as it isn't clear exactly what is going on. I don't know if it would be helped with #2773, but in any case, it seems the fallback mode shouldn't apply already closed revisions, since they are most likely to be on autoland anyway.

La0:
Fixes #2827

Depends #2842 & https://github.com/mozilla/libmozdata/pull/258

### DIFF
--- a/bot/code_review_bot/analysis.py
+++ b/bot/code_review_bot/analysis.py
@@ -51,6 +51,13 @@ class RevisionBuild(PhabricatorBuild):
     def __repr__(self):
         return str(self)
 
+    def load_patches_stack(self, phabricator_api):
+        """
+        Load the stack of patches from Phabricator API
+        """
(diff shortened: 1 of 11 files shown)


suhaibmujahid (approved):
Thank you! LGTM!

Archaeopteryx (approved):

PR #2851, opened 2025-06-27
Remove try syntax

La0:
Followup of #2842

As try syntax [has been deprecated](https://bugzilla.mozilla.org/show_bug.cgi?id=1648719), we can remove the 2 try mode  support (json & syntax) to only use json.


Archaeopteryx (dismissed):

La0:
This is rebased on top of master

PR #2886, opened 2025-07-18
Apply patches from Phabricator in a smarter way

vrigal:
Ref. https://github.com/mozilla/code-review/issues/2887

The hglib client only supports the similarity parameter (and not the `config` option): https://repo.mercurial-scm.org/python-hglib/file/tip/hglib/client.py#l959. Therefore, I could not implement the fall back command in case the command fails. It should be possible with a raw call to `hg`, but I suppose we want to avoid that.

I did not added the `--no-commit` option as Bastien told me it would work only on try.

### DIFF
--- a/bot/code_review_bot/mercurial.py
+++ b/bot/code_review_bot/mercurial.py
@@ -643,7 +643,6 @@ class MercurialWorker:
 
         except Exception as e:
             logger.warn("Failed to process diff", error=e, build=build)
-            raise e
             return (
                 "fail:general",
                 build,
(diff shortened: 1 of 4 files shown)


La0 (approved):

cgsheeh:
> @cgsheeh is this still the way Lando does it?
> 
> > Lando does an `hg import --no-commit -s 95` and if that fails, it then runs a `hg import --no-commit --config ui.patch=patch -s 95`, which falls back to using the patch command

Yes, this still is how Lando [applies patches to Mercurial repos](https://github.com/mozilla-conduit/lando/blob/main/src/lando/main/scm/hg.py#L235).

La0:
You should be able to rebase on top of https://github.com/mozilla/code-review/pull/2941

La0 (approved):
LGTM, especially with a unit test trigerring both imports

PR #2902, opened 2025-08-04
warn but do not error if GITHUB_API_TOKEN not set in secret or configuration file

Archaeopteryx:


marco-c:
Why not making it mandatory?

PR #3003, opened 2025-10-08
Use default revision instead of tip as fallback

### LINKED ISSUE #2994: Fallback to "default" instead of "tip"
`tip` is an ambigious reference: it can be the current tip of any branch in the repository. In many cases on hg.mozilla.org, this ends up being a `tags` branch, which are forked once from an old cycle and never updated.

`default` is ìwhat we want here, as it will always be the tip of the default branch.

La0:
Closes #2994

### DIFF
--- a/bot/tests/conftest.py
+++ b/bot/tests/conftest.py
@@ -1030,7 +1030,7 @@ def build_repository(tmpdir, name):
     readme = tmpdir.join(name, "README.md")
     readme.write("Hello World")
     repo.add(str(readme.realpath()).encode("utf-8"))
-    repo.branch(name=b"central", force=True)
+    repo.branch(name=b"default", force=True)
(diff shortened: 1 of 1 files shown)


marco-c:
There's also "tip" in `bot/code_review_bot/revisions.py` which we probably need to replace with "default"

bot/tests/test_analysis.py
```diff
@@ -107,7 +107,7 @@ def mock_hgrun(cmd):
             "robustcheckout",
             b"--purge",
             f"--sharebase={tmpdir}/mozilla-central-shared".encode(),
-            b"--branch=tip",
+            b"--branch=default",
```
marco-c, line 110: @cgsheeh does this look OK?

PR #3044, opened 2025-10-27
Use taskcluster index for all modes

### LINKED ISSUE #3024: Index new bot tasks
Code review bot tasks can create any Taskcluster route under `project.relman.<ENV>.code-review`, but we only currently write `index.project.relman.<ENV>.code-review.latest` for the new analysis mode.

The `Workflow.index` method must be updated to support the new mode, and create meaningful routes allowing a code-review admin to track down both analysis & triggering tasks using Phabricator IDs, and git/hg revisions

La0:
Closes #3024

This PR uses the existing Taskcluster index integration in all 3 modes so that every task generated by the bot will get indexed.

### DIFF
--- a/bot/tests/test_index.py
+++ b/bot/tests/test_index.py
@@ -123,7 +123,6 @@ def test_index_autoland(
         "try_task_id": "remoteTryTask",
         "url": None,
     }
-    assert calls[0][0][1]["data"] == payload
     assert all([c[0][1]["data"] == payload for c in calls])
 
 
@@ -183,7 +182,6 @@ def test_index_phabricator(
         "try_task_id": "remoteTryTask",
         "url": "https://phabricator.test/D51",
     }
-    assert calls[0][0][1]["data"] == payload
     assert all([c[0][1]["data"] == payload for c in calls])
 
 
@@ -248,5 +246,4 @@ def test_index_from_try(
         "try_task_id": "remoteTryTask",
         "url": "https://phabricator.test/D51",
     }
-    assert calls[0][0][1]["data"] == payload
     assert all([c[0][1]["data"] == payload for c in calls])


bot/code_review_bot/revisions.py
```diff
@@ -134,12 +134,33 @@ def __init__(
...
+        if self.diff_phid:
+            out.append(f"phabricator.diffphid.{self.diff_phid}")
+
+        # Revision indexes
+        if self.head_repository and self.head_changeset:
+            repo = repo_slug(self.head_repository)
+            out.append(f"head_repo.{repo}.{self.head_changeset}")
+        if self.base_repository and self.base_changeset:
+            repo = repo_slug(self.base_repository)
+            out.append(f"base_repo.{repo}.{self.base_changeset}")
```
Archaeopteryx: Will this be useful? Multiple different Phabricator revisions will update this index with their task ID.

Archaeopteryx: This modifies `namespaces`, a list of strings which gets used [here](https://github.com/mozilla/code-review/blob/b48a3e6bd380a854c6c6477873b89186241c8fee/bot/code_review_bot/workflow.py#L462).

```
            self.index_service.insertTask(
                namespace,
                {
                    "taskId": settings.taskcluster.task_id,
                    "rank": 0,
                    "data": payload,
                    "expires": stringDate(now + timedelta(days=TASKCLUSTER_INDEX_TTL)),
                },
            )
```

If `base_repo.{repo}.{self.base_changeset}` links to the last task which used it as a base revision, is there a good use for it?

bot/tests/test_index.py
```diff
@@ -0,0 +1,252 @@
...
+        "source": "try",
+        "state": "unit-test",
+        "target_repository": "https://hg.mozilla.org/mozilla-unified",
+        "title": "Changeset deadbeef123 (https://hg.mozilla.org/integration/autoland)",
+        "try_group_id": "remoteTryGroup",
+        "try_task_id": "remoteTryTask",
+        "url": None,
+    }
+    assert calls[0][0][1]["data"] == payload
+    assert all([c[0][1]["data"] == payload for c in calls])
```
Archaeopteryx: Based on line 97, calls is not empty. Should line 127 stand alone and line 126 be dropped because 127 covers it? Similar below.

Archaeopteryx (approved):

PR #3107, opened 2025-12-09
Local port of the libmozevent.phabricator module

vrigal:
Port of `libmozevent.phabricator` into code review bot directly. This version includes  https://github.com/mozilla/libmozevent/pull/109, that has been [reverted](https://github.com/mozilla/libmozevent/pull/115) in `libmozevent` latest version 1.1.34

### DIFF
--- a/bot/code_review_bot/sources/phabricator.py
+++ b/bot/code_review_bot/sources/phabricator.py
@@ -5,6 +5,7 @@
 import collections
 import enum
 import time
+from datetime import datetime, timedelta
 
 import structlog
 from libmozdata.phabricator import PhabricatorAPI
@@ -16,6 +17,7 @@ class PhabricatorBuildState(enum.Enum):
     Queued = 1
     Secured = 2
     Public = 3
+    Expired = 4
 
 
 class PhabricatorBuild:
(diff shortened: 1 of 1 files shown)


bot/code_review_bot/sources/phabricator.py
```diff
@@ -0,0 +1,227 @@
...
+
+        # Then we can check on the expiry date
+        date_created = build.diff.get("dateCreated")
+        if not date_created:
+            logger.warn("No creation date found", build=str(build))
+            return False
+
+        logger.info("Found diff creation date", build=str(build), created=date_created)
+
+        return datetime.now() - datetime.fromtimestamp(date_created) > self.build_expiry
```
Archaeopteryx: Should this be `datetime.now(timezone.utc)` (with `utcnow()` deprecated) to avoid different behavior when executed locally?

La0: Probably but this could be a followup ? (as this is a 1-to-1 port of the libmozevent code).

BTW This snippet would then probably be "promoted" to a method shared by both Phabricator & Github sources

PR #3143, opened 2026-01-14
Support github pull requests on the bot, backend and frontend

La0:


La0:
This is the base MR for github support with bot + backend changes in the same place.

PR #3223, opened 2026-02-27
If patch fails to apply, warn it is likely to fail to land and provide instructions how to resolve (#3222)

### LINKED ISSUE #3222: add better message when patch fails to apply, currently 'abort: uncommited changes'
Code review bot failed to run for https://phabricator.services.mozilla.com/D285267 which had been developed standalone but afterwards had been configured to have a new parent revision: https://phabricator.services.mozilla.com/D285157
Between the initial base revisions of those two changes, https://phabricator.services.mozilla.com/D284354 had landed and modified code adjacent to a change in D285267.

`moz-phab patch D285267` applied D28517 first and used it as base for D285267 which effectively rebased it. Because of the changes from D284354 missing, the patch failed to apply.

Code review bot posted to Phabricator's D285267 two comments:

```
WARNING: The code review bot failed to apply your patch.

abort: uncommitted changes
```

Changes:
# Warn directly the change will likely fail to land
# Suggest a rebase in case the developer has manually edited the parent revision.

Archaeopteryx:


marco-c:
This would be a nice improvement for merge conflict resolution (even if it catches a small minority of cases: patches that have merge conflicts when posted to Phab).
We should report it as a warning to Lando too.

CC @cgsheeh @zzzeid @shtrom

zzzeid:
> This would be a nice improvement for merge conflict resolution (even if it catches a small minority of cases: patches that have merge conflicts when posted to Phab). We should report it as a warning to Lando too.

Couple of thoughts:
- This could be possible to do with small modifications on the code-review-bot side, since it currently posts a warning that static analysis is in progress, and later deletes that warning. Instead of deleting it, it could modify the text, or delete it and create a new one that shows the intended warning (no Lando change required here).
- I suspect that code-review-bot does not currently apply patches based on remote `HEAD`, which is what Lando currently does. This may cause some discrepancies in what can and can't be merged.

marco-c:
I think we might not need this anymore, with merge conflict detection built in Phabricator and the upcoming fix for https://bugzilla.mozilla.org/show_bug.cgi?id=2068637.

PR #3252, opened 2026-03-11
Backend support for github provider

vrigal:
This PR is the first part that was split off from all the work done in #3175


vrigal:
@Archaeopteryx, if you have time to review this PR with the backend changes it would be greatly appreciated.
It contains all the changes required by the bot to publish Revision, Diff and Issues from a Github PR (See #3255).

backend/code_review_backend/issues/api.py
```diff
@@ -166,7 +172,7 @@ def get_queryset(self):
         if query is not None:
             search_query = (
                 Q(id__icontains=query)
-                | Q(revision__phabricator_id__icontains=query)
+                | Q(revision__provider_id__icontains=query)
```
Archaeopteryx, line 175: Does this rely on the different providers having different formats for `provider_id`: integer vs. hash?

backend/code_review_backend/issues/serializers.py
```diff
@@ -52,9 +52,12 @@ def to_internal_value(self, url):
             except (TypeError, ValueError):
                 self.fail("invalid")
                 return
+        # Github repositories uses a path like <owner>/<slug>, so we need to replace the
+        # inner slash to have a slug usable within URLs (e.g. to list issues on a repo).
+        slug = parsed.path.lstrip("/").replace("/", "-")
```
Archaeopteryx: `mozilla-firefox/browser` and `mozilla/firefox-browser` would return the same. What do you think about using `+` instead?

backend/code_review_backend/issues/api.py
vrigal: Indeed, it would either match a Phabricator integer (as a string) or hash for github (both stored as a `varchar(40)`).

backend/code_review_backend/issues/serializers.py
vrigal: I also made the change in #3254

Archaeopteryx (approved):

PR #3423, opened 2026-06-12
Bump docker image artifact expiry to a year

jcristau:
If we don't push to the production branch for 6 months the image expires and hooks start to fail.  Try to make that less likely.

Also update the task description to not pretend it's using docker in docker.


Archaeopteryx (approved):

marco-c:
@Archaeopteryx is this good to merge?

PR #3476, opened 2026-07-03
Avoid harbormaster crash when publishing a link

La0:
By looking at recent [erroneous tasks from Taskcluster prod environement](https://code-review.moz.tools/#/tasks) (filtering by state `error`), I noticed that most of the issues come from an exception raised early in some analysis tasks that fail to publish a link on harbormaster.

The Phabricator API raises a 400 about duplicates (which is OK), but the bot immediately crashes.

I simply added a try/except clause around the publication, and a unit test to cover the happy path (no existing tests  on that method).

### DIFF
--- a/bot/code_review_bot/workflow.py
+++ b/bot/code_review_bot/workflow.py
@@ -7,7 +7,7 @@ from datetime import datetime, timedelta
 from itertools import groupby
 
 import structlog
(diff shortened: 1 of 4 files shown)


marco-c:
Duplicate of #3444?

Archaeopteryx (dismissed):

Archaeopteryx (approved):

PR #3488, opened 2026-07-07
task type selector: set DefaultTask as last pick

Archaeopteryx:


Archaeopteryx:
Issue is the artifact list used by `DefaultTask.matches` can be empty and better scoped matches by task names should be used first to avoid this.

bot/code_review_bot/workflow.py
```diff
@@ -691,9 +691,7 @@ def build_task(self, task_status):
             raise Exception(f"Cannot read task name {task_id}")
 
         # Default format is used first when the correct artifact is available
```
La0:
```suggestion
        # Specific tasks detections are enabled first, default format is used as a fallback
```

PR #3490, opened 2026-07-09
Bug 2048352 - Migrate code-review bot to push to Github 

nahadji:
Adds a Git push path to the code-review bot so the analysis workflow can push
patch stacks to a GitHub repository instead of hg.mozilla.org/try
([Bug 2048352](https://bugzilla.mozilla.org/show_bug.cgi?id=2048352)).

The change is additive: the Mercurial path is untouched, and the Git backend is
selected per repository via a new optional `repo_type` field in the
repositories configuration (defaults to `hg`, so existing configurations keep
working unchanged).

### What's included
- `GitRepository`: applies a Phabricator patch stack as Git commits (preserving
  authorship), writes the same `try_task_config.json`, and force-pushes to a
  configured branch on the remote. No Lando `git2hg` lookup (the base is
  already a Git hash); missing bases fall back to the default revision and are
  reported to Phabricator like the Mercurial path does.
- `GitWorker`: mirrors `MercurialWorker`, minus the treestatus wait (no "try"
  tree to gate on). Eligible push errors are retried with exponential backoff.
- Backend selection in `start_analysis` based on `repo_type`.
- Failure publication: `fail:git` results are reported to Phabricator and Lando.
- Auth: pushes are authenticated over HTTPS with a short-lived GitHub App
  installation token (generated per run with simple-github, restricted to the
  try repository, never written to disk). The bot reads GITHUB_APP_ID and
  GITHUB_APP_PRIVKEY from the runtime configuration secret.

### Verified
- Unit tests (`tests/test_git.py`, extended `tests/test_phabricator_analysis.py`);
  self-contained (local git repos, no Mercurial environment needed).
- The push pipeline was exercised for real against
  `mozilla-releng/staging-firefox` with a write deploy key (patch stack applied,
  `try_task_config.json` committed, branch pushed and verified, then deleted).

### DIFF
--- a/bot/code_review_bot/cli.py
+++ b/bot/code_review_bot/cli.py
@@ -56,6 +56,11 @@ def parse_cli():
         default=os.environ.get("TASKCLUSTER_SECRET"),
     )
     parser.add_argument(
+        "--git-ssh-key-secret",
+        help="Taskcluster secret holding the SSH deploy key used to push to Git try repositories",
+        default=os.environ.get("GIT_SSH_KEY_SECRET"),
+    )
+    parser.add_argument(
         "--mercurial-repository",
         help="Optional path to a up-to-date mercurial repository matching the analyzed revision.\n"
         "Reduce the time required to read updated files, i.e. to compute the unique hash of multiple issues.\n"
@@ -116,6 +121,12 @@ def main():
     # Setup libmozdata configuration
     setup_libmozdata("code-review-bot")
 
+    # Load the dedicated Git deploy key from its own secret when configured
+    git_ssh_key = None
+    if args.git_ssh_key_secret:
+        secret = taskcluster.get_service("secrets").get(args.git_ssh_key_secret)
+        git_ssh_key = secret["secret"]["ssh_privkey"]
+
     # Setup settings before stats
     settings.setup(
         taskcluster.secrets["APP_CHANNEL"],
@@ -124,6 +135,7 @@ def main():
         taskcluster.secrets["ssh_key"],
         args.mercurial_repository,
         args.github_repository,
+        git_ssh_key=git_ssh_key,
     )
 
     # Setup statistics
--- a/bot/code_review_bot/cli.py
+++ b/bot/code_review_bot/cli.py
@@ -56,11 +56,6 @@ def parse_cli():
         default=os.environ.get("TASKCLUSTER_SECRET"),
     )
     parser.add_argument(
-        "--git-ssh-key-secret",
-        help="Taskcluster secret holding the SSH deploy key used to push to Git try repositories",
-        default=os.environ.get("GIT_SSH_KEY_SECRET"),
-    )
-    parser.add_argument(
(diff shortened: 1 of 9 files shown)


bot/code_review_bot/cli.py
```diff
@@ -124,6 +126,8 @@ def main():
         taskcluster.secrets["ssh_key"],
         args.mercurial_repository,
         args.github_repository,
+        github_app_id=taskcluster.secrets["GITHUB_APP_ID"],
```
La0: It would be nicer to create a dedicated `GITHUB` maaping in the conf with the following parts:
- `app_id`
- `app_privkey`

This would allow adding more configuration later on (as the needs for Github will certainly grow in that code base).

Also you need to document these new options there : https://github.com/mozilla/code-review/blob/master/docs/configuration.md

bot/code_review_bot/config.py
```diff
@@ -24,7 +24,10 @@
 )
 RepositoryConf = collections.namedtuple(
     "RepositoryConf",
-    "name, try_name, url, try_url, decision_env_prefix, ssh_user",
+    "name, try_name, url, try_url, decision_env_prefix, ssh_user, repo_type",
+    # repo_type is optional and defaults to Mercurial so existing repository
```
La0, line 28: Please document that in https://github.com/mozilla/code-review/blob/master/docs/configuration.md

bot/code_review_bot/workflow.py
```diff
@@ -295,21 +297,39 @@ def start_analysis(self, revision):
...
+                    "try_url": base_conf.try_url,
+                    # GitHub App credentials to generate short-lived push tokens
+                    "github_app_id": settings.github_app_id,
+                    "github_app_privkey": settings.github_app_privkey,
+                },
+                cache_root=settings.git_cache,
+            )
+            worker = GitWorker()
+        else:
+            repository = Repository(
```
La0: You should rename `Repository` to `MercurialRepository` to avoid any confusion

La0 (changes requested):
There is a lot of code duplication for the new GitWorker & GitRepository classes, you should create a base class that shares as most code as possible between both git & hg implementations, then inherit from these base classes into specific implementations.

The ideal way to do that is to make a first PR which splits the existing `Worker` & `Repository` into a base class + specific mercurial implementation, then create another PR on top that implements the git part (what you did here).

Things like the configuration in `__init__`, patch normalization, the whole workflow to apply a build (albeit specific git/mercurial applications which can be further extracted), try commit build, skippable commit detection, retry detection, the overall worker `run` can be shared across classes.

You should end up with something like that:

```mermaid
flowchart TD

  base_worker[BaseWorker] --> hg_worker[MercurialWorker]

  base_worker[BaseWorker] --> git_worker[GitWorker]


  base_repo[BaseRepository] --> hg_repo[MercurialRepository]

  base_repo[BaseRepository] --> git_repo[GitRepository]
```

PR #3504, opened 2026-07-15
Make source-test-doc-upload tasks check both issues and upload files.

Standard8:
The `source-test-doc-upload` builders are now publishing an `issues.json` artifact, so `DefaultTask` is being run to process the issues, however, this means that the `DocUploadTask` is no longer being run, so reviews don't get the links to the documents.

PR #3488 partially addresses this by making sure the `DocUploadTask` runs, but then the `DefaultTask` won't be run - so we'll revert to the previous state of uploading files, but not showing the details of an published issues.

This PR builds on top of #3488 (both are needed), and changes the `DocUploadTask` to inherit from `DefaultTask`, so that we'll check both the issues and upload the files if possible.

I went with the inheritance approach, as this seemed better than spawning multiple tasks.

### DIFF
--- a/bot/code_review_bot/tasks/default.py
+++ b/bot/code_review_bot/tasks/default.py
@@ -54,13 +54,22 @@ class DefaultTask(AnalysisTask):
     https://github.com/mozilla/code-review/blob/master/docs/analysis_format.md
     """
 
-    artifacts = ["public/code-review/issues.json"]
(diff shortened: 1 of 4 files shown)


Standard8:
Just to be clear, the unit test is failing because #3488 needs merging first.

Archaeopteryx (approved):

PR #3529, opened 2026-07-27
Bug 2048352 - Split the push workflow into VCS agnostic base classes

nahadji:
Extract the version control agnostic parts of the mercurial push workflow into BaseRepository and BaseWorker (new vcs module): building and applying a stack of Phabricator patches, committing try_task_config.json, the skippable files detection, the push retry loop and the worker output handling.

The Mercurial specific parts stay in MercurialRepository (renamed from Repository to avoid confusion with the upcoming Git implementation) and MercurialWorker: robustcheckout and batch clones, revision lookups with the Lando git2hg fallback, patch application through hg import, ssh push and the treestatus wait.

No behavior change: this prepares the introduction of a Git implementation of the same workflow as subclasses of the new bases (First step of the review feedback on #3490).

### DIFF
--- a/bot/code_review_bot/vcs.py
+++ b/bot/code_review_bot/vcs.py
@@ -1,0 +1,355 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+import os
+import time
+
+import rs_parsepatch
+import structlog
+from libmozdata.phabricator import PhabricatorPatch
+
+from code_review_bot.sources.phabricator import PhabricatorBuild
+
+logger = structlog.get_logger(__name__)
+
+TREEHERDER_URL = "https://treeherder.mozilla.org/#/jobs?repo={}&revision={}"
+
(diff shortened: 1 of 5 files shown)


bot/code_review_bot/vcs.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # Apply patches to the latest revision when `True`.
+        self.use_latest_revision = config.get("use_latest_revision", False)
+
+        self._repo = None
+
+    def __str__(self):
+        return self.name
+
+    def clone(self):
```
La0, line 60: Please add a docstring on every non-implemented base methods: il will be easier later on (or in the future for next developers to work on that).

La0 (commented):
The  Worker `handle_build` method could output a dedicated class instance in the future, so it's easier to process across various implementations (don't do it there, it's already a great base to build upon).

bot/code_review_bot/vcs.py
nahadji: Added Docstrings for the non-implemented base methods in commit 33883bb

La0 (approved):

marco-c:
@La0 can we merge this?

Archaeopteryx (approved):

PR #3578, opened 2026-08-20
add support for running bugbug-selected tasks for autoland-based reviewbot pushes

bhearsum:
A large percentage of failures on autoland are caught by bugbug-selected tasks. I'd like to experiment with having bugbug-selected tasks run on reviewbot pushes to see if it helps developers notice and fix such problems earlier.

Enabling this mostly requires tweaks to the parameters; it will also require a change on the Gecko side to ensure all of the right tasks are in target tasks (otherwise there's nothing bugbug to choose from).

We shouldn't do this until reviewbot reports failures back to phab without waiting for the entire task group to be completed, otherwise reporting, eg: lint failures will be delayed.

### DIFF
--- a/bot/code_review_bot/config.py
+++ b/bot/code_review_bot/config.py
@@ -67,6 +67,12 @@ class Settings:
         # Indexed by their Phabricator ID
         self.user_blacklist = {}
 
+        # bugbug test configuration
+        self.bugbug_enabled_repositories = ["firefox-autoland"]
+        # Because it's unclear how much load this will add, this is being rolled out gradually.
+        self.bugbug_enabled_percent = 0.1
+        self.bugbug_optimize_strategy = "gecko_taskgraph.optimize:tryselect.bugbug_reduced_manifests_config_selection_medium"
+
         # Always cleanup at the end of the execution
         atexit.register(self.cleanup)
         # caching the versions of the app
--- a/bot/code_review_bot/vcs.py
+++ b/bot/code_review_bot/vcs.py
@@ -4,12 +4,14 @@
 
 import json
 import os
+import random
 import time
 
 import rs_parsepatch
 import structlog
 from libmozdata.phabricator import PhabricatorPatch
 
+from code_review_bot.config import settings
 from code_review_bot.sources.phabricator import PhabricatorBuild
 
 logger = structlog.get_logger(__name__)
@@ -20,6 +22,13 @@ TREEHERDER_URL = "https://treeherder.mozilla.org/#/jobs?repo={}&revision={}"
 MAX_PUSH_RETRIES = 4
 # Wait successive exponential delays: 6sec, 36sec, 3.6min, 21.6min
 PUSH_RETRY_EXPONENTIAL_DELAY = 6
+
+
+def bugbug_enabled(repository_name):
+    if repository_name not in settings.bugbug_enabled_repositories:
+        return False
+
+    return random.random() < settings.bugbug_enabled_percent
 
 
 class RetryNeeded(Exception):
@@ -176,6 +185,13 @@ class BaseRepository:
                 "phabricator_diff": build.target_phid,
             },
         }
+
+        if bugbug_enabled(self.name):
+            config["parameters"]["test_manifest_loader"] = "bugbug"
+            config["parameters"]["optimize_strategies"] = (
+                settings.bugbug_optimize_strategy
+            )
+
         diff_phid = build.stack[-1].phid
 
         if build.revision_url:
--- a/bot/tests/test_mercurial.py
+++ b/bot/tests/test_mercurial.py
@@ -7,11 +7,13 @@ import time
 from unittest.mock import MagicMock
 
 import hglib
+import pytest
 import responses
 from conftest import MockBuild
 from libmozdata.phabricator import PhabricatorPatch
 
 from code_review_bot import mercurial
+from code_review_bot.config import settings
 
 
 class STDOutputMock:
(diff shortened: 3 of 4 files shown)


bot/code_review_bot/config.py
```diff
@@ -67,6 +67,12 @@ def __init__(self):
         # Indexed by their Phabricator ID
         self.user_blacklist = {}
 
+        # bugbug test configuration
+        self.bugbug_enabled_repositories = ["firefox-autoland"]
```
bhearsum, line 71: I realize that a bunch of config for bugbug is in tc secrets. I wasn't sure if that was helpful or desired here; I can make that adjustment if needed.

bot/code_review_bot/config.py
```diff
@@ -67,6 +67,12 @@ def __init__(self):
...
+        # Because it's unclear how much load this will add, this is being rolled out gradually.
+        self.bugbug_enabled_percent = 0.1
+        self.bugbug_optimize_strategy = "gecko_taskgraph.optimize:tryselect.bugbug_reduced_manifests_config_selection_medium"
```
bhearsum, line 74: In an ideal world, I'd want to use the exact same strategy as autoland. In reality that doesn't seem possible because it uses a long composite strategy that takes into account push id, backstops, and other things. I'm not sure if what I have here is the best one; I'm happy to be corrected on this.

bot/code_review_bot/config.py
```diff
@@ -67,6 +67,12 @@ def __init__(self):
         # Indexed by their Phabricator ID
         self.user_blacklist = {}
 
+        # bugbug test configuration
+        self.bugbug_enabled_repositories = ["firefox-autoland"]
+        # Because it's unclear how much load this will add, this is being rolled out gradually.
+        self.bugbug_enabled_percent = 0.1
```
Archaeopteryx, line 73: Incorrect variable name

bot/code_review_bot/config.py
Archaeopteryx: On autoland, the builds for various operating systems get scheduled. On try, the fewer builds the better, and they better are fast ones and without a small machine pool to run one. By default Linux opt should run, debug if the penalty for slower tests is acceptable. macOS and Windows shall get test tasks if the changes are expected to affect operating systems differently.

Archaeopteryx:
There are plans to meet with Bastien next week and have dkl set up a separate build plan for build + test which are not linting. For linting, reporting after each failed task should rarely report more than one failed task. For test tasks or even multiple builds, the results need to be bundled and if possible deduplicated.

@marco-c will be able to guide about the scheduling strategy.

Without builds scheduled by default, will `run-on-projects: built-projects` schedule the tasks?

Archaeopteryx:
This also needs to cancel the already running tasks if a new version of the patch is submitted.

bhearsum:
> There are plans to meet with Bastien next week and have dkl set up a separate build plan for build + test which are not linting. For linting, reporting after each failed task should rarely report more than one failed task. For test tasks or even multiple builds, the results need to be bundled and if possible deduplicated.

Thanks; I wasn't aware of these plans.

> Without builds scheduled by default, will `run-on-projects: built-projects` schedule the tasks?

They'll get pulled in when tests are chosen, or when bugbug decides to schedule them explicitly. https://treeherder.mozilla.org/jobs?repo=try&revision=fa79adf8be1d73ff7848525b4cc96fb3f1f106ef is an example I hacked together that didn't need this (it inserts the necessary parameter changes directly)

bhearsum:
> build plan for build + test which are not linting

Does this mean a separate reviewbot push would happen? That would solve the reporting delay concerns for sure.

bhearsum:
> This also needs to cancel the already running tasks if a new version of the patch is submitted.

https://github.com/mozilla/code-review/pull/3590 for this; I think it's a pretty independent piece that can land ahead of this.

bot/code_review_bot/config.py
bhearsum: Thank all makes sense to me.

@marco-c - do you have any comment on which precise optimization strategy would be preferable here?

bhearsum:
This probably requires some tweaks still, but I'm going to move it out of drafts to make it clear that it's ready for further review.

bot/code_review_bot/config.py
marco-c: @bhearsum I'm thinking we might want to introduce a new one that runs a smaller subset of tests (and maybe only tests that are never intermittent?), otherwise we'd almost duplicate the costs of autoland.

bhearsum: Sure, that seems like a good starting point. Do we have that already? If not, how would I add it?

marco-c: Looks like we don't have a bugbug_reduced_manifests_config_selection_high in addition to bugbug_reduced_manifests_config_selection_medium, so we could introduce that as a starting point (https://searchfox.org/firefox-main/rev/bde88e0509310cea9d16d473d94d7d0f48632df2/taskcluster/gecko_taskgraph/optimize/__init__.py#212 and https://searchfox.org/firefox-main/rev/bde88e0509310cea9d16d473d94d7d0f48632df2/taskcluster/gecko_taskgraph/optimize/bugbug.py#67-70).

For intermittents, we'd need to add something either in bugbug or in taskgraph to query intermittency data (I think @fqueze has a task that is publishing an artifact with this data) and skip tests that are intermittent.

bot/code_review_bot/config.py
bhearsum: What's incorrect about it?

bhearsum:
This will end up getting reworked a fair amount; moving back to drafts.

PR #3590, opened 2026-08-24
feat: cancel in-progress reviewbot pushes before new ones are started

bhearsum:
This was originally motivated by https://github.com/mozilla/code-review/pull/3578, but it's valid and worthwhile on its own.

Aside from the very roundabout way we have to find task group ids, this is pretty straightforward: simply find and cancel all previous task groups for the revision. This avoids doing work for something that's already stale.

If someone knows of a way to find the task group id of the reviewbot push other than the roundabout way it's happening here, I'd be very happy to switch to it. I could not find any way to pull it (or even the treeherder link) through the API directly; the only thing that seems to available there is the task id of the initial code review task.

In addition to the unit tests, I managed to run code review bot locally as some sort of integration test. I didn't have it actually cancel any tasks, but I _think_ I've done enough to make this landable. Ideally we can test in a non-prod environment before production.

### DIFF
--- a/bot/code_review_bot/workflow.py
+++ b/bot/code_review_bot/workflow.py
@@ -2,6 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import base64
+import json
+import re
 import time
 from datetime import datetime, timedelta
 from itertools import groupby
@@ -44,6 +47,13 @@ logger = structlog.get_logger(__name__)
 
 TASKCLUSTER_NAMESPACE = "project.relman.{channel}.code-review.{name}"
 TASKCLUSTER_INDEX_TTL = 7  # in days
+
+DECISION_TASK_ROUTE = "gecko.v2.{repo}.revision.{revision}.taskgraph.decision"
+PUBLICATION_LOG_ARTIFACT = "public/logs/live.log"
+TREEHERDER_LINK_REGEX = re.compile(
+    rb"treeherder\.mozilla\.org/(?:#/)?jobs\?repo=(?P<repo>[\w-]+)"
+    rb"&revision=(?P<revision>[0-9a-f]{12,40})"
+)
 
 
 class Workflow:
@@ -347,6 +357,11 @@ class Workflow:
         worker = MercurialWorker()
         output = worker.run(repository, build)
 
+        # Cancel any in-progress tasks from an earlier update
+        # This is done after pushing to try to avoid delaying runs of the
+        # new tasks.
+        self.cancel_previous(revision)
+
         # Update index when the patch has been pushed to try
         self.index(revision, state="pushed_to_try")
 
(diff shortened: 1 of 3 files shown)


marco-c:
A downside of this is that the build on Phabricator would show up as pending, maybe there's some sort of "cancelled" result we could upload to Phab?

marco-c:
Regarding Ben's question, @La0 can you think of another way to retrieve the task group?

bhearsum:
> A downside of this is that the build on Phabricator would show up as pending, maybe there's some sort of "cancelled" result we could upload to Phab?

It would, but only on the previous Build, which probably doesn't matter...?

In any case, I'll see if I can update the state in Phab, as it would be the Proper (tm) think to do.

La0:
> Regarding Ben's question, @La0 can you think of another way to retrieve the task group?

Currently i think it's the only way.

But it would not be complicated to store the task group in the backend [next to `Diff.review_task_id`](https://github.com/mozilla/code-review/blob/master/backend/code_review_backend/issues/models.py#L129)

Then you could fetch that info through the [revision_diffs_list endpoint](https://api.code-review.moz.tools/docs/)

bhearsum:
I managed to get a proper test done locally. I'm pretty darn confident this will work in production. You can see it in action in this revision: https://phabricator.services.mozilla.com/D320141. The previous diff had its buildable cancelled: https://phabricator.services.mozilla.com/D320141?id=1363741, and the try push for it was also cancelled: https://treeherder.mozilla.org/jobs?repo=try&revision=27b4f18dc9a63fa448de26b353d9ac1dbfb5b28c.

Latest update here contains a couple of notable changes:
* Support for aborting previous buildables
* Seal task groups prior to cancelling (otherwise cancelling doesn't work)

As part of testing I also discovered that there is a theoretical scenario where cancellation is missed when a second update is made to the revision in short order, and the previous try push doesn't have a decision task yet. I'm choosing not to care about this, because this scenario is very unlikely once try pushes fire off of github events in the not too distant future. It would be good to fix still, but in _that_ world it's much less critical.

> > Regarding Ben's question, @La0 can you think of another way to retrieve the task group?
> 
> Currently i think it's the only way.
> 
> But it would not be complicated to store the task group in the backend [next to `Diff.review_task_id`](https://github.com/mozilla/code-review/blob/master/backend/code_review_backend/issues/models.py#L129)
> 
> Then you could fetch that info through the [revision_diffs_list endpoint](https://api.code-review.moz.tools/docs/)

I didn't implement this. I can as part of this if necessary, but I would prefer to do it in a follow-up later, if that's OK.

bot/code_review_bot/workflow.py
```diff
@@ -465,6 +480,213 @@ def publish(self, revision, issues, task_failures, notices, reviewers):
...
+        """
+
+        # In order to cancel tasks from a previous push we need the task group id
+        # that the tasks ran under (which is the same as the decision task for that push).
+        # The only way to retrieve this through the Phabricator API is by a long series
+        # of requests:
+        # * Find all of the prior Build Targets (via builds, via buildables)
+        # * Find the log for each Build Target, which will contain a taskId of Code Review task
+        # * Pull the Code Review task log to fetch the treeherder link with the revision pushed to Try in it
+        # * Look up the decision task id in the task index via the revision
```
marco-c, line 496: One way to at least cut this short would be to add the decision task ID to the publication task, so you could read the publication task definition instead of having to parse the logs.

marco-c (approved):
LGTM, with the caveat that this only works with Phabricator and Mercurial, and doesn't with GitHub PRs and git or Phabricator and git (which is not supported yet but will soon).

bot/code_review_bot/workflow.py
bhearsum: Is the publication task the one that posts the results back to phab? If so, that means we'll be getting the information too late to matter? (If it's the one that runs to make the try push, that's too early...we don't have the decision task id yet.)

PR #3626, opened 2026-09-11
Bug 2068637 - Use first public parent as ReviewBot base fallback

olliehcrook:


bot/code_review_bot/mercurial.py
```diff
@@ -202,12 +202,17 @@ def repo(self):
 
         return self._repo
 
-    def get_mercurial_base_hash(self, revision):
+    def get_mercurial_base_hash(
+        self, revision: str | None, revision_type: str = "base revision"
```
marco-c: To simplify things, I think it's fine to keep the same wording (just "revision") in both cases, so let's remove the `revision_type` parameter.

marco-c (changes requested):
Looks good overall, just a minor simplification suggestion

bot/code_review_bot/mercurial.py
```diff
@@ -202,12 +202,15 @@ def repo(self):
...
 
-    def get_mercurial_base_hash(self, revision):
+    def get_mercurial_base_hash(self, revision: str | None) -> str | None:
         """A revision may reference to a Git commit hash instead of Mercurial one.
         The revision can either be a 40 characters full hash or its first 12 characters (short hash).
         A Lando API enables to "convert" the Git hash to a Mercurial hash that can
         be found in the local repository, whatever its length.
         """
+        if not revision:
+            return None
```
marco-c: nit: if the only caller is line 263, no one is using the function with a "None" value, so we could remove this check and change the type definition to be "str" instead of "str | None"

marco-c (dismissed):

marco-c (approved):

PR #3639, opened 2026-09-11
Parallelize issue hashes computation

### LINKED ISSUE #3638: We are spending a lot of time computing issue hashes
E.g. in https://firefoxci.taskcluster-artifacts.net/ErAfZ8x-Tymdm4O5QCsU8A/0/public/logs/live_backing.log it's more than a minute (where the total runtime of the task is 3 minutes and 30 seconds.

We could calculate issue hashes in parallel instead of sequentially.

marco-c:
Fixes #3638


bot/code_review_bot/__init__.py
```diff
@@ -186,47 +186,40 @@ def in_patch(self):
...
         return self.revision.in_touched_files(self)
 
-    @cached_property
+    @property
+    def is_autogenerated(self):
+        """
+        An autogenerated file resides in the build directory that has the
+        format `obj-x86_64-pc-linux-gnu`
+        """
+        return "obj-" in self.path
```
La0, line 195: Isn't that a bit generic and maybe you need to check that any part **starts with** `obj-` ?


```suggestion
        return any([part.startwith("obj-") for part in self.path.parts])
```

bot/code_review_bot/revisions/phabricator.py
```diff
@@ -79,6 +79,11 @@ def __init__(
         # Patch analysis
         self.patch = patch
 
+    @property
+    def local_repository(self):
```
La0, line 83: Nit: could be named `local_repository_path` to explicitly state this does not provide a `Repository` instance

La0 (commented):
Not much to say on the refactoring, I'll test on a revision and report on perf.

bot/code_review_bot/__init__.py
marco-c: It was already like this, I didn't change it, but I think we can change it to be `startswith("obj-")`. I'll do it in a separate follow-up PR though since it's pre-existing.

La0:
On a clean repo (full checkout) on that branch:

```
real	8m11,888s
user	4m44,746s
sys	2m2,043s
```

Second run (no checkout):

```
real	0m30,358s
user	0m6,765s
sys	0m4,917s
```

On current master 7bbadf23a7a4e82e2ed2cd59b55ed94b7b4d6f3a , warm cache (no checkout):

```
real	0m27,762s
user	0m7,051s
sys	0m4,654s
```

That test revision is pretty inconclusive as it had very few hashes to calc (almost all issues silent). I'll retry with a revision with more publishable issues

marco-c:
The case I saw was in https://firefoxci.taskcluster-artifacts.net/ErAfZ8x-Tymdm4O5QCsU8A/0/public/logs/live_backing.log, you could try with that revision.

marco-c:
I've run some benchmarks myself, evicting the filesystem cache and adding a minimal 1ms delay when reading files (since gcp is slower than local disk).

It is faster, but it was not that slow actually. What made it slow at the time was the huge number of logged messages, which is already gone thanks to #3625.

It goes from around 4 seconds to 0.5 seconds. Is it worth it? Not sure.

PR #3643, opened 2026-09-14
Misc optimizations

marco-c:
Review commit by commit.


bot/code_review_bot/config.py
```diff
@@ -17,6 +17,9 @@
 REPO_MOZILLA_CENTRAL = "https://hg.mozilla.org/mozilla-central"
 REPO_AUTOLAND = "https://hg.mozilla.org/integration/autoland"
 
+# Number of concurrent requests made to remote services (backend, Taskcluster)
+MAX_PARALLEL_REQUESTS = 8
```
La0: That could be configured through Taskcluster to be able to control per-instance as testing & prod are not using the same dynos for the backend

bot/code_review_bot/workflow.py
```diff
@@ -235,10 +236,18 @@ def _build_tasks(tasks):
             nb=len(supported_tasks),
         )
 
-        # Load all the artifacts and potential issues
+        # Load all the artifacts in parallel, as each task download is independent
+        with ThreadPoolExecutor(max_workers=MAX_PARALLEL_REQUESTS) as executor:
```
La0: That could use a different setting as it targets Taskcluster which probably has more resources than the code review backend

La0 (dismissed):

bot/code_review_bot/config.py
marco-c: Done.

bot/code_review_bot/workflow.py
marco-c: Done.

La0 (approved):

PR #3667, opened 2026-09-18
feat: wire in support for additional types of tests when pushing to Try

bhearsum:
This is a functionally equivalent refactor that is a useful precursor to https://bugzilla.mozilla.org/show_bug.cgi?id=2065267. It's best looked at commit-by-commit (each one has some additional details in the commit message). The high level overview is that it allows different code paths to be taken in analysis and publication mode that will allow us to push to try with different parameters, and analyze and publish results in different ways.

Perhaps the most notable part here is the adjustment of where `build_target_phid` is fetched from. This is not _strictly_ necessary as part of this refactor, but when we add a build&test try push, we won't have a `code-review` task that we can fetch this from, and it's already available in the decision task parameters so I figured we may as well be consistent with where we fetch it from.

This has been tested locally against phabricator-dev. I managed to kick off [a sane looking try push](https://treeherder.mozilla.org/jobs?repo=try&revision=649dc0e7044ce05965bbdc50c48827fbf8108688&selectedTaskRun=fVTicTt2SAifgW4N5l25bw.0), and when I re-ran publication mode against a previous code review bot run, it produced identical results.


Archaeopteryx:
Looks reasonable to me. Bastien for second set of eyes on it.

